PR #36, opened 2015-11-19
add error serializer, standardized json errors, and tokens controller

scarfdestroyer:
Added @begedin error formatting and their required helpers for login with standard credentials. Also added tokens_controller, error_serializer, and the respective tests.


begedin:
Why was "awaiting review" removed fromt this one? Seems good to me :+1:

PR #37, opened 2015-11-20
Password reset

scarfdestroyer:
added password reset functionality from @JoshSmith Cook Academy

We still need to add:
- [ ] Mailer tests
- [ ] Mailer
- [ ] Web host URL in ENV to send in mailer


begedin:
Looks like this one was branched off of #36 (my guess is, due to using common code), so I recommend merging #36 before reviewing and merging this one.

app/controllers/passwords_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      render json: user
+    else
+      render_could_not_reset_password_error
+    end
+  end
+
+  private
+
+    def find_user_by_confirmation_token
+      Clearance.configuration.user_model.find_by_confirmation_token(params[:id].to_s)
```
joshsmith: Change this to `params[:token]` and send the token in the body. Remove it from the URL. Method should stay `PUT` for now.

scarfdestroyer:
I changed the password reset logic from the passwords_controller to the users_controler. The creation of the confirmation code and email still resides in the passwords_controller but the actual reset route is under `POST /users/reset_password`. This was to remove the necessity of putting the confirmation token in the url in order to receive it as an `:id` in the `PUT passwords` route.

spec/requests/api/passwords_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      post "#{host}/passwords", {
+        user: {
+          email: "not-existing-user@mail.com"
+        }
+      }
+      expect(last_response.status).to eq 422
+      expect(json.errors.email).to include "doesn't exist"
+    end
+  end
+end
```
joshsmith: My feeling is that like `reset_password`, this now belongs under a `users/forgot_password` route. What do you think?

joshsmith: @scarfdestroyer let me know what you think on this.

joshsmith:
@scarfdestroyer if you can add in the extra functionality listed in the description, that would get this mergeable.

joshsmith:
@scarfdestroyer actually going to merge to work this into something else I'm working on. You can do this as a separate PR.

Issue #27, opened 2015-07-14
Add GET /user endpoint


joshsmith:
We need a `GET` /user endpoint so we can have user profiles.

scarfdestroyer:
https://github.com/code-corps/code-corps-api/pull/38

joshsmith:
Completed by #38.

PR #38, opened 2015-11-20
add user routes and controller action

scarfdestroyer:
https://github.com/code-corps/code-corps-api/issues/27


joshsmith:
This doesn't look like it's using a `UserSerializer` or checking for the right JSON data as a result.

PR #40, opened 2015-11-23
Add authenticated user endpoint

joshsmith:
This adds an authenticated user endpoint for returning a differently serialized user for the person. It's similar to `/me` endpoints we've used in the past.


begedin:
@JoshSmith The problem was doorkeeper. The version we were using (2.1.0) has an actual bug, involving the method we use to render the error. The version in ground-game is 2.2.0, so there was no such issue.

I fixed it by deleting the `gemfile.lock` and rerunning the `bundle` command. This updated doorkeeper to `3.0.1`, which makes me think we should consider doing the same with ground-game at some point, since we're obviously using out of date gems.

I used the gem `'pry-stack_explorer'` to figure this out and I left it in, since it's a pretty useful gem. It allows us to explore the call stack when prying the code. We an use `show-stack` to print the stack in a list with frame_id. We can then use `frame frame_id` to jump to that specific frame of the stack and see what the arguments were.

Issue #32, opened 2015-07-14
Add Twitter to user model


joshsmith:
A user should be able to enter their Twitter handle.

Issue #31, opened 2015-07-14
Add website to User model


joshsmith:
A user should be able to link to their website. Add standard validations for URLs.

PR #41, opened 2015-11-24
Add twitter, website and biography attributes to user

### LINKED ISSUE #30: Add biographical section to user model
A user should be able to post a short text-only biography.

begedin:
Will close #30, #31 and #32 when merged.

Most resources I found online suggest using URI::regexp to test for validity of an URL; but that expression requires the protocol to be specified, so I rolled my own, based on various others I've found. The specs in the user model ensure which formats we support or explicitly do not support.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -128,7 +128,7 @@ ActiveRecord::Schema.define(version: 20151124130007) do
     t.string   "username"
     t.boolean  "admin",                          default: false, null: false
     t.text     "website"
-    t.text     "twitter"
+    t.string   "twitter"
     t.text     "biography"
   end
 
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -108,6 +108,9 @@
     t.string   "remember_token",     limit: 128,                 null: false
     t.string   "username"
     t.boolean  "admin",                          default: false, null: false
+    t.string   "website"
+    t.string   "twitter"
+    t.string   "biography"
```
joshsmith: I think anywhere that we don't know something will be short, we need to default to using `text`. There seems to be no performance hit to Postgres, so the two are essentially the same.

joshsmith:
@scarfdestroyer can you make these changes? You should be able to just change the original migration to `text` without issue.

joshsmith:
Don't edit the schema directly!

Issue #11, opened 2015-07-14
Create GET /project endpoint


joshsmith:
We need a `GET /project` endpoint for an individual project.

Issue #9, opened 2015-07-14
Add Project model


joshsmith:
A `Project` should have a:
- [x] `title:string`
- [x] `description:string`
- [ ] and an icon

begedin:
#35 adds title and description. Since icon needs additional handling, I delayed adding that until we merge #35 and also add project edpoints.

@JoshSmith Will the icon be an image we upload, or something small enough so we can save it as a base64 string?

joshsmith:
@begedin it will be an image a user uploads.

joshsmith:
Icon can be implemented in #44.

Issue #18, opened 2015-07-14
Add pagination to post endpoints


joshsmith:
We should be able to paginate the posts since there may be quite a number them.

begedin:
Is there a fixed length to a page, or will we be specifying it in the request? The later gives more flexibility to me, and shouldn't be difficult to implement.

joshsmith:
I think we'd specifying it in the request, and there probably needs to be meta information for the pagination to work (number of pages, etc).

PR #43, opened 2015-11-24
Add GET endpoint for individual and list of projects

begedin:
Closes ##10, #11 when merged. Possibly #9 to, but depends.

I'm marking this as awaiting review because it's mergeable, but there are some things that could be done either via this PR, or in a separate PR
- [ ] Project should also have an icon. Since that involves uploads, I'm guessing we want to handle that somehow, via s3/cloudfront/just local uploads? This feels like something worthy of a PR of its own.
- [ ] Not sure which relationships to list by default and if I should include records for those relationships. Right now, the serializer only exposes the attributes (title, description, id).

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,0 +1,10 @@
+class ProjectsController < ApplicationController
+  def index
+    render json: Project.all
+  end
+
+  def show
+    project = Project.find(params[:id])
+    render json: project
+  end
+end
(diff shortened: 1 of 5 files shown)


app/controllers/projects_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ProjectsController < ApplicationController
+  def index
+    render json: Project.all
```
joshsmith, line 3: We'll definitely want to add some form of pagination to this. This is not sustainable in its current form.

This is #18.

joshsmith:
I think icon as separate PR. Issue #44

Issue #48, opened 2015-11-25
Reset and forgot password errors do not follow JSON API specs


begedin:
This is what they look like at the moment:

``` Ruby
def render_no_such_email_error
  render json: {errors: {email: ["doesn't exist"]}}, status: 422
end

def render_could_not_reset_password_error
  render json: {errors: {password: ["couldn't be reset"]}}, status: 422
end
```

I believe these should be replaced either by initializing a `ActiveModel::Errors` object (so we just serialize them as validation errors), or we should define our own custom errors and have the serializer serialize those.

joshsmith:
:+1:

Issue #47, opened 2015-11-25
Validation errors do not follow JSON API specs


begedin:

joshsmith:
Closed by #49.

PR #49, opened 2015-11-25
Patch users endpoint

### LINKED ISSUE #33: Add PATCH endpoint for users
We should be able to update user information.

begedin:
Closes #33, #46, #47
- In `rescue_from CanCan::AccessDenied`, we now render the error using `ErrorSerializer`
- We now render validation errors using `ErrorSerializer`
- Only admin users can edit other users using `"PATCH /users/:id"` - **I'm thinking we might not actually want this endpoint**
- Current user can edit themselves using `"PATCH /users/me"`
- Only `[:website, :biography,:twitter]` can be edited.
- The following endpoints now follow JSON API specs, both in requests they accept and responses they return
  - `"POST /users"`
  - `"POST /users/forgot_password"`
  - `"POST /users/reset_password"`
- The exception of the previous statement is that the password endpoints return errors in a format that does not follow the JSON API spec. For that purpose, I've created #48 and I think it should be handled separately, in one of the ways described in the issue.


joshsmith:
What's the reasoning behind not wanting `PATCH /users/:id`? I'm not sure how I feel about this one, just want to hear your arguments behind it.

Maybe create a separate issue for this @begedin.

spec/support/helpers/request_helpers.rb
```diff
@@ -23,7 +23,20 @@ def authenticated_put(path, args, token)
...
   end
+
+  def convert_to_json_api_hash(hash, type)
+    data = {}
+    data[:id] = hash[:id] if hash[:id].present?
+    data[:type] = type
+    data[:attributes] = hash.except(:id)
+
+    { data: data }
+  end
```
joshsmith, line 41: I would consider renaming this method `json_api_params_for(type, hash)`, which reads a little better: `json_api_params_for("users", {})`.

joshsmith, line 41: And I changed it.

Issue #50, opened 2015-11-25
Work out access rights for post and comment creation


begedin:
- When should the user be allowed to create posts for a project?
  - Is it `post_type` specific? Role-specific? Project type specific? We need to flesh this out.
- When should the user be allowed to create comments for a post? 
  - Some overlap with posts here, but also differences, I think.

joshsmith:
I think this is a combination of role inside a project and the post type.

I think comments on posts will always be allowed, regardless of the post type or relationship to a project.

Issues and ideas can be posted by anyone.

Progress posts and tasks can only be created by contributors to a project. This should encourage a higher level of ownership.

Also, the poster can close any post they created, but posts can be closed by project admins at will, though not contributors.

In this sense, the hierarchy is:
- Owner
- Admins
- Contributors
- Everyone

begedin:
This makes sense, but we there's no point in partially implementing it, until we have contributors. Blocked by #12

joshsmith:
This was done for posts in #103 but we need to do it for comments. Creating a separate issue though.

Issue #24, opened 2015-07-14
Add endpoint to create comments


joshsmith:
We should `POST` to create comments on a post.

PR #51, opened 2015-11-25
Create posts comments api

### LINKED ISSUE #21: Add endpoint for creating a post
We need to `POST` to create a `Post`.

begedin:
Will close #21, #24 when merged
# Description

This is still a work in progress. Since I made validation and access errors follow the JSON API in #49, I decided to add "pending" statements to specs in this one until #49 is merged. Otherwise, I would have to do repeat work as well as additional work resolving conflicts once both are ready for merge.

Outside of that, here's what we have:
- We can GET individual and multiple posts, no pagination yet, since there's a separate issue for that
- We can create posts
  - `title` is required
  - `body` is not required
  - `type` was changed to `post_type`, since `type` is a reserved name
  - `type` is permitted but has a default set to "task"
  - `status` is permitted as a parameter, but is ignored. Instead, we set it to "open" by default.
  - `project_id` is required as a parameter
  - `user_id` is required as a parameter, but is auto-set to the current user
  - Anyone who is signed in can create posts - this is a very basic ability and needs to be improved on, in a separate issue, I think - #50
- We can't get individual or multiple comments for now. I'm thinking they can be included with posts.
  - Once we have paging, we should be able to filter comments by post_id and then retrieve individual pages.
- We can create comments
  - `body` is required
  - `post_id` is required
  - `user_id` is required, but is auto-set to the current user


joshsmith:
#49 is merged.

begedin:
Since #49 is merged, and pending specs are fixed, this should be mergeable as well.

joshsmith:
I'm not sure I understand why this isn't passing, but maybe you can fix it. Good to merge, though I would consider making `body` in `Post` required.

JoshSmith:
simulator screen shot nov 24 2015 6 07 53 pm

begedin:
@JoshSmith Regarding making the body of post required, I was going with the same approach as github, where you can create an issue with just a title, no body.

begedin:
@JoshSmith If you could quickly review this, now that the tests pass, I could merge it (or add the post body requirement, then merge), since it's blocking #60

joshsmith:
Looks good to me, but I definitely think a body is a good idea. I don't like Github's approach personally. Leads to over-long titles for things.

PR #56, opened 2015-11-26
Fix password errors

### LINKED ISSUE #48: Reset and forgot password errors do not follow JSON API specs

begedin:
Closes #48 
# Description

I'm initializing a custom `ActiveModel::Errors` object to achieve this and then render it through our already established pipeline. Since this  class requires an instanced model to be passed in as an attribute, I'm using a newly initialized `User`.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -37,10 +37,11 @@ class UsersController < ApplicationController
 
   def forgot_password
     user = User.find_by(email: forgot_password_params[:email])
+
     if user && user.forgot_password!
       render json: user
     else
-      render_no_such_email_error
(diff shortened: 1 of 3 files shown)


joshsmith:
:+1:

Issue #58, opened 2015-11-26
Should the admin be able to create and delete user_skills for other users through the "/user_skills" endpoint


begedin:
@JoshSmith  See title

joshsmith:
I don't see this happening all that often.

begedin:
Closing this, since it was answered and there's nothing else to do about it for now.

Issue #57, opened 2015-11-26
When creating a user_skill, if the associated skill does not exist, should we be able to create it?


begedin:
@JoshSmith

What is the user interface like in this case? Do we select a skill from a dropfdown/auto-complete when adding it to our profile, or are we also able to create a completely new skill?

joshsmith:
For now it'll be some sort of autocomplete functionality, or selecting from a list. But I don't think we should allow users to create a bunch of random skills. We'll seed it initially and only later consider opening that up.

begedin:
In that case, the `POST /user_skills` endpoint has this behavior covered. It requires the skill to be specified in the request and returns a validation error otherwise.

However, we do need to expose a `GET /skills` in this case, if we want an autocomplete #64.

begedin:
Closing this in favor of #64

PR #59, opened 2015-11-26
Skills and categories

### LINKED ISSUE #6: Add skills with categories
We want a `Skill` model that includes a `title:string` and a `category`. For example, `Ruby` is in the `Development` category.

begedin:
Closes #6, #7, #8
# Implemented behavior
- A skill has a `title` (required), a`description` (not required) and a `category` (required)
- A category has a `title` (required)
- A `user_skill` has a `user` and a `skill` (both required) and validates that the (`user_id`,`skill_id`) pairs are unique, so we can't create a duplicate
- We have an endpoint for creating `user_skills`
  - It sets the `user` of the `user_skill` to the `current_user` automatically. There is currently no way to assign it manually. We can do it in another PR if we need that feature - #58
  - It requires the `skill` of the `user_skill` to be set to an already existing skill. It cannot create a new one through the request at the moment. If we need that behavior, we can implement it separately - #57 
- We have an endpoint for deleting 'user_skills'
  - It requires the user of the skill to be the current_user. If we want the admin to do it for other users, we can do a separate PR - #58
  - It deletes the `user_skill` only. It doesn't touch the `user` or the `skill`.
- There is no PUT/PATCH endpoint, since there is nothing to edit other than the relationships
# TODO
- [x] Rename `Category` to `SkillCategory`


begedin:
@JoshSmith This is mergeable, except for the question of the PUT/PATCH endpoint. I've marked the related issue #8 with a question label.

joshsmith:
Do you think `Category` should be a `SkillCategory`?

begedin:
Might be a good idea, depending on what else we might have eventually. It's more descriptive in any case.

begedin:
@JoshSmith Performed the rename. Is this good to go otherwise?

PR #60, opened 2015-11-26
Post likes

### LINKED ISSUE #22: Add a PostLike or similar model
We want the ability to "like" posts, so we need a way for users to like them. There should probably be a counter cache column for this as well.

begedin:
Closes #22 when done
# Implementation

I could've reimplemented it here to just move on but I decided to just implement everything else I could instead, since there's a limited amount of time anyway.
- current user can like a post (by creating a `PostLike`
- current user can unlike a post (by deleting a `PostLike`
- there is no way for an admin to like or unlike a post in someone else's name. Unlike user_skills, I don't think there's any reason for that here.
- api validates the request, there are validation errors if
  - the post with the  specified id doesn't exist
  - no post id is specified in the request
- there is a counter cache column for post likes on the post model


spec/requests/api/post_likes_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+          included_users = json.included.select{ |i| i.type == "users" }
+          expect(included_users.count).to eq 1
+        end
+
+        it "includes post in the response" do
+          expect(json.included).not_to be_nil
+
+          included_users = json.included.select{ |i| i.type == "posts" }
+          expect(included_users.count).to eq 1
```
joshsmith, line 56: This should be `included_posts`.

joshsmith:
We have minimal tests for counter cache for a `Post`'s likes, but we should really be testing that more fully.

joshsmith:
Added some tests in here on the model and serializer level and fixed the migration to set a `default: 0` for the counter cache column.

PR #61, opened 2015-11-27
Expose skills on show user

### LINKED ISSUE #29: Add skills to user endpoint
We should be able to see a user's skills on thier profile.

begedin:
Closes #29 

Branches off of #59, so that one needs to be merged first to make the changeset here clear

**Ready for review, but blocked due to #59 not being merged yet**
# Implementation
- `GET /users/:id` endpoint didn't have any specs yet, so I added that
- `UserSerializer` now has a `has_many` relationship to `skills`
  - There is a spec on the serializer which covers this.
  - I'm not showing the `user_skills` relationship here, since I feel there's no need for that for the purposes of showing it on a public profile. Just the `skills` the user has "through" `user_skills`
- `UsersController#show` now renders the user json with the `incude: ["skills"]` option active.
  - There's an API spec covering this


joshsmith:
Have you checked this for N+1 queries? Would be nice to have tests for that, actually, if it's not altogether too crazy to do.

My guess would be that this has a lot of N+1 problems. You can look at CookAcademy's API for examples of this exact issue in action, and how I solved it there.

begedin:
I'm expecting a decent amount of them at this point. I'll create a task to handle existing ones and set up the gem to detect new ones in the future.

PR #63, opened 2015-11-28
add validations for username format

scarfdestroyer:
Added failing specs.


begedin:
Merged in latest "develop" branch into this one and fixed the specs so they don't fail anymore. Not sure if anything else needs to be added.

PR #66, opened 2015-11-30
Posts pagination

### LINKED ISSUE #18: Add pagination to post endpoints

begedin:
Closes #18 
# Implementation

Two new methods defined on the application controller level:

``` Ruby

def page_size
  params.fetch(:page, {}).fetch(:size, 10).to_i
end

def page_number
  params.fetch(:page, {}).fetch(:number, 1).to_i
end

```

We then use the `Kaminari` gem, which is supported by AMS out of the box and makes AMS add a `links` object to the `json` response.

``` Ruby
def index
  posts = Post.page(page_number).per(page_size)
  render json: posts
end
```

It's as straightforward as that.

Kaminari's page numbers are 1-based, not 0-based, so the default page is 1, while the default page size is 10. We might want to move this to a config of some sort, but Kaminari also supports several other ways to configure defaults, so I delayed that until there's a need and we figure out which approach is best for us.
# Response

Response will now include a `links` object, in the format 

``` JavaScript
{
  self: "link to current page",
  first: "link to first page"
  prev: "link to previous page, doesn't exist if current is first",
  next: "link to next page, doesn't exist if current is last",
  last: "link to last page"
}
```

It will also include a `meta` object, with the format

``` Javascript
{
  total_pages: "total number of pages in the current page size",
  total_records: "total number of records in general",
  current_page: "page number specified in the request for the current response",
  page_size: "page size specified in the request for the current response"
}
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,9 +23,11 @@ gem 'aws-sdk'
 
 gem 'sidekiq'
 
+gem 'kaminari'
+
 group :development, :test do
   gem 'dotenv-rails'
-  
+
   gem 'pry-rails'
   gem 'pry'
   gem 'pry-nav'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -95,6 +95,9 @@ GEM
     jmespath (1.1.3)
     json (1.8.3)
     jwt (1.5.2)
+    kaminari (0.16.3)
+      actionpack (>= 3.0.0)
+      activesupport (>= 3.0.0)
     loofah (2.0.3)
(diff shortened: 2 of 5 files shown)


joshsmith:
Awesome. :+1: 

Looks like pagination is going to be difficult for now on the Ember side (no syntactic sugar): https://github.com/emberjs/data/issues/2905

joshsmith:
Maybe could use this as well: https://github.com/mharris717/ember-cli-pagination

PR #68, opened 2015-11-30
Add markdown to posts

joshsmith:
This adds a `markdown` field to the `Post`. A `Post` will take this in on `create` and in a `before_validation` hook it will render the markdown into HTML using `Redcarpet`, and then save that as the `body` which gets rendered out from the API.

### DIFF
--- a/spec/requests/api/posts_spec.rb
+++ b/spec/requests/api/posts_spec.rb
@@ -115,7 +115,7 @@ describe "Posts API" do
 
       it "does not require a 'post_type' to be specified" do
         params = { data: { type: "posts",
-          attributes: { title: "Post title", markdown: "<p>Post body</p>\n" },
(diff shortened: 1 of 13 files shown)


spec/requests/api/posts_spec.rb
```diff
@@ -115,7 +115,7 @@
 
       it "does not require a 'post_type' to be specified" do
         params = { data: { type: "posts",
-          attributes: { title: "Post title", body: "Post body" },
+          attributes: { title: "Post title", markdown: "<p>Post body</p>\n" },
```
begedin: You're specifying HTML here. Is it intentional?

joshsmith: Oh, no. I should be specifying markdown.

begedin:
Looks good, other than the HTML thing.

Issue #67, opened 2015-11-30
Switch CanCan to Pundit


begedin:
https://github.com/elabs/pundit

The existing model needs to be switched, so we can start implementing the Contributor model in #12

Issue #13, opened 2015-07-14
Add endpoint for joining a project


joshsmith:
This should be an endpoint for contributors.

begedin:
Blocked by #12

begedin:
Based on how #12 is being implemented, we could go in multiple directions with this one. If we want simple RESTful routes, then we'd need

# First approach
- `POST /contributors` - This one creates a new Contributor record, with `project` set to provided project, `user` set to `current_user` and `status` set to "pending". It can be accessible by any logged in user.
- `PATCH /contributors` . This one takes an existing `Contributor` record and sets its status to some other value. At first, it can be used by an admin or owner to approve a pending `Contributor` record. Later, it can also be used to change contribution levels (the value of `status`) for other contributors as well.

This first approach gives more flexibility, but also puts some responsibility on the client. When the user clicks on "apply for a contributor" (or whatever they click on), the client app will need to know this means creating a new contributor record with `project` set to the project currently being viewed.

# Second approach

We have routes such as
- `POST /projects/:id/apply_for_collaboration` (or some other similar route)
- `POST /contributors/:id/approve,
- `POST /contributors/:id/promote`
- etc.

The actions are more atomic this way and the routes are more verbose and descriptive, but I'm not really sure the client will be able to be dumber this way, especially an ember client.

Due to that, I'll go with the first approach for now, but I can shift to the second if necessary.

begedin:
@JoshSmith  I believe this was handled by #70.  We now have a `POST /contributors` route, which enables joining a project.

However, I believe we should change this route so it's nested under the project - `POST /:member_slug/:projet_slug/contributors` to maintain consistency. I'm creating a new issue to address this and closing this one - #107

PR #70, opened 2015-12-01
Add contributor model

### LINKED ISSUE #12: Add Contributor model
We need a `Contributor` model that relates a `User` to a `Project`, with some metadata like their status in the project (admin, etc). We should look to GitHub for a good set of statuses.

begedin:
Closes #12 , #13

Branches off of #67, so that one needs to be merged first.
# Implementation

A `Contributor` belongs to a `User` and a `Project`, both required. It also has a required `status` enum, which initially defaults to `"pending"`. Values for `status` are `[:pending, :collaborator, :admin, :owner]`.

`GET /contributors` returns a list of contributors for the specified project. It **requires the project to be specified**. It follows the JSON API convention of putting the filter parameters inside a filter object.

`POST /contributors` requires a project to be specified. It creates a contributor record with project assigned to the specified project, user assigned to the current user and status assigned to "pending".

`PATCH /contributors/:id` requires a status to be specified. It allows or disallows changing the status to the new value based on what the new value is, what the old value was and that contributor role the current user has on the project.
- admin can promote contributors from "pending" to "collaborator", but they cannot do anything else.
- owners can do the same, and can additionally promote "collaborator" to "admin"
- owners can also demote "admin" to "collaborator"
# TODO
- Specs for the `ContributorPolicy`
- I'm not sure our access rights are correct. Does the above sound about right?
- Should I add a DELETE endpoint? That would be the way to kick people out of the project, I think. Who can do it? Owners only, or admins to?


joshsmith:
@begedin I think admins actually have same abilities as the owner, but the owner can do things like transfer the project to another user/organization. Teams or projects will have a limited number of admins. Not going to be something that happens often, but will be good to spread the work around for overwhelmed project owners.

joshsmith:
I think a DELETE endpoint makes sense. Owners and admins both, as in https://github.com/code-corps/code-corps-api/pull/70#issuecomment-161038242.

joshsmith:
@scarfdestroyer I would do things in this order:
- [ ] Specs for `ContributorPolicy`
- [ ] Add a `DELETE` endpoint that's authed to owners and admins
- [ ] Change admins to have same abilities as owner (for now)
- [ ] Create an issue for us to make it so only admins can transfer ownership of a project to another user/organization

joshsmith:
@begedin this is in progress. I would work on other things as you can without getting held up by this.

joshsmith:
Still in progress, but can be expanded on by whomever wants to take it on.

app/policies/contributor_policy.rb
```diff
@@ -0,0 +1,58 @@
...
+  def index?
+    return true
+  end
+
+  def create?
+    return unless @contributor_for_user.present?
+
+    return true if @contributor.status == "owner" && current_user_is_owner_of_project?
+    return true if current_user_is_at_least_admin_on_project?
+    return true if @contributor.status == "pending"
```
begedin: Are we sure this is the way we want to have it?

The way I see it, any user should be allowed to apply to a project. This is just access rights we're defining here.

Of course, a user should not be able to apply if an application already exists, but that's handled by validation (validating uniqueness of a user scoped to a project) and should return a validation error, not an access right error.

In this definition
- Line 14 means current user will not be able to create a new application unless they already have an application made. There's also a bug here. Need to call `contributor_for_user`, not the `@contributor_for_user`. This way, `@contributor_for_user` will never be set.
- Line 16 means current user can create an owner contributor record if they are also the owner. That makes sense. It means an owner is allowed to transfer ownership to someone else.
- Line 17 means only an admin or an owner can create a contributor record.
- Line 18 means anyone can create a `pending` contributor, but line 14 ensures that only user who are already contributors can do it.

The way I see it, `create?` should return `@user.present? && @contributor.status == "pending"`. This means any logged in user can apply to be a contributor. Only contributor records with "pending" status should be allowed to be created directly.

app/policies/contributor_policy.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    if current_user_is_owner_of_project?
+      # Promote a collaborator to admin
+      return true if contributor.status_was == "collaborator" and contributor.admin?
+      # Demote an admin to collaborator
+      return true if contributor.status_was == "admin" and contributor.collaborator?
+    end
+
+    return false
```
begedin, line 38: I like the definition of `update?`. It makes sense to me.

Additionally, if we want to remove a user from being a collaborator, that should go into a delete action, I'm thinking for `delete?` admins can delete "pending" and "collaborator", while owners can also delete "admin".

Finally, for ownership transfer, I don't think that should be done on a `Contributor` level at all. Instead, it should go to the project level.

begedin:
@scarfdestroyer @JoshSmith I added a comment on how I think the `create?` policy should work. I think we need to discuss it.

I also added a policy for delete? and specs which cover `delete?`, `update?` and `index?`.

I'm thinking there may be confusion with the `create?` policy and access rights, due to the `admin` property on the user model. I don't think it really relates to the contributor model in any way, or am I wrong in this?

At best, the `User.admin?` property should relate to all policies on all objects in the same way - it should grant all rights.

scarfdestroyer:
The Project model didn't have the `has_many :contributors` so I added that, the counter cache and the migration for the `contributors_count` to store the cache. Also I originally did this on another branch but then realized I branched it from this one so it had all of those failing tests but develop doesn't have the contributor model at all so I closed/deleted it and moved it here.

PR #73, opened 2015-12-01
Add post user mentions

joshsmith:
This is the initial work around `post_user_mentions`.

We still need to:
- [ ] Add a state machine to `Post` for `draft` and `published`
- [ ] Add some sort of `edited` state (don't think this belongs in state machine, but just maybe a `bool` for determining whether to show that a post has been edited)
- [x] Create a serializer for these mentions
- [x] Create a convenience method for the indices so they're wrapped inside an array, i.e. `[1, 8]` instead of `start_index` and `end_index`, for use with the API
- [x] Write API specs to see that the `post_user_mentions` are included
- [x] Ensure there are no n+1 queries
- [ ] We want to make notifications polymorphic, so that things are `notifiable`
- [ ] Add a model for the `PostUserNotifications` for actually notifying and clearing notifications for a user around a given `PostUserMention`
- [ ] Ensure duplicate `PostUserNotifications` are not created even if a user gets re-mentioned multiple times
- [x] Write test to ensure underscores don’t get parsed out in usernames in markdown


joshsmith:
@begedin I'm still torn on whether to be doing the @mention manipulation on front-end or not. On the one hand, Twitter appears to prefer it be done on the front-end. On the other hand, Github appears to do it via what they call an HTML pipeline. https://github.com/jch/html-pipeline

We could apply a `MarkdownFilter` and a `MentionFilter` and all sorts of other things.

I don't, though, see how the filtering could be done on-the-fly in a way that's at all performant, especially in comments. While the filtering chaining may make sense to do at every other level, I still worry that it's not appropriate in the case of our mentions.

If we render and _then_ store, then we have the issue of URLs getting stale.

Could definitely use thoughts on this.

joshsmith:
I think the problems with on-the-fly are compounded by needing to query for all these users just to render the appropriate URL, to avoid the staleness issue.

joshsmith:
I hate to rely on engineering team size and userbase, but it makes me think that Twitter is doing it the way they do for a reason, that is maybe superior to Github. But still give me your (as unbiased as possible) opinion.

joshsmith:
@begedin pinging you again to get your thoughts on this.

begedin:
@JoshSmith I didn't consider the pipeline approach. At the most basic level, we could even just transform the HTML as we're fetching the post/comments, on the API side.

I'm really not sure what's best. 
- Client side transforming requires a smarter client, which is something we wanted to avoid, but it's the least amount of work for the server.
  - The server fetches post, with user_mentions and comments included (with comments also including mentions of their own) and all mentions including the user, returns all of it in one go.
  - Client goes through mentions and does the text replacement.
  - Server only renders html once, at post/comment save. Least amount of stress on the server.
  - No staleness issues
- Server does the replacement. The client can be completely dumb.
  - Server needs to re-render post/comment for every request and do the replacement before returning ti to the client
  - No staleness issues
  - Lots of work for the server
  - No work for the client
- Server does the mention insertion on post/comment save
  - Only does it once, not on every request, so less work. Only slightly more than first approach
  - Staleness issues
  - No work for the client

Approach 2 would be ideal if we have multiple different clients, but I'm not sure how it will perform. Approach 1 is the safer one to go with. Approach 3 means user would have to re-save the comment/body to eliminate stale mentions. If that works, then we could go with that.

joshsmith:
@begedin I don't think `with_includes` is failable in the way we want it to be.

joshsmith:
This needs to be wrapped up and I'll move the other tasks into something separate. For now, it's doing the job that was set out to be done.

joshsmith:
First couple tasks are moved to here: https://github.com/code-corps/code-corps-api/issues/98

joshsmith:
Last few tasks are here: https://github.com/code-corps/code-corps-api/issues/99

scarfdestroyer:
:+1:  :100:

PR #75, opened 2015-12-02
Facebook authentication (with no special actions)

### LINKED ISSUE #2: Add ability to sign up with Facebook to API

begedin:
Closes #2 
# Implementation

I've used most of the code from ground-game/fieldthebern. Among other things, I had to add support for serialization of several new errors which weren't yet being serialized in this project.

Outside of this, there is nothing else happening here.

When creating a user, we do the same as creating a regular user, except `facebook_id` and `facebook_access_token` are permitted parameters now, meaning we store them as User fields as well.

When logging in through Facebook, we take the `facebook_access_token`, use it to get the `facebook_id`, then use that to find the user in the database.  

In both cases, **nothing else is happening yet**. In this flow, the client is the one that fill in user details from Facebook upon user creation, so that part isn't handled by the API. We don't have user photos or social features yet either, so that part doesn't need to be processed.
## New RSpec matcher

I also added a new RSpec matcher which allows us to simply check if the proper serializer is being used to create the json response for our request. This saves us from having to check each individual field, for which there should be no need, since we're already testing each serializer individually.

We can use this matcher to significantly reduce the amount of code in our request specs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,8 @@ gem 'sidekiq'
 
 gem 'kaminari'
 
+gem 'koala'
+
 gem 'redcarpet'
 
 group :development, :test do
@@ -49,6 +51,9 @@ end
 group :test do
   gem 'factory_girl_rails', '~> 4.0'
   gem 'shoulda-matchers'
+
+  gem 'vcr'
+  gem 'webmock'
 
   gem 'oauth2'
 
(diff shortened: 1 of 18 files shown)


begedin:
@JoshSmith The specs are failing due to Facebook environment variables missing from our build. I used my own generic ones which I usually use for development locally.

As usual, we need

```
FACEBOOK_APP_ID=
FACEBOOK_APP_SECRET=
FACEBOOK_REDIRECT_URL=

```

PR #76, opened 2015-12-02
Slug user organization endpoint

### LINKED ISSUE #39: Add a slugged endpoint for users/organizations
We need to implement something like what's implemented here: http://blog.arkency.com/2014/01/short-urls-for-every-route-in-your-rails-app/

You would hit `api.domain.com/:slug` and render either a `User` or an `Organization`. Since there are more users than organizations (i.e. a likelier hit), we should find users first, and then query for organizations before rendering a 404.

begedin:
Closes #39 

Still very much a work in progress
# Implementation
- Only `GET /:slug` gets redirected to `organizations#show` or `users#show`. `GET /:id` is not matched with anything.
- Using `friendly_id` to achieve all of this.
- If there's an organization and user with the same slug, organization takes precedence
- Direct endpoints `GET /users/:id` and `GET /organizations/:id` still exist.
  - [ ] Should these be removed?
  - [ ] I added specs to ensure `GET /users/:slug` and `GET /organizations/:slug` also work. Should these be removed as well?
  - [ ] If any of these should be removed, I need to update the `requests/api/users_spec.rb` and `requests/api/organizations_spec.rb` so they send GET requests to the base, friendly slug route. In my opinion, we should keep it as is, though. A client app in the future might prefer to use direct routes, and it doesn't hurt us at all, as long as we have the routing spec I defined here.


begedin:
@JoshSmith This is marged as "question"; but if we aren't removing the existing routes, then you can consider it as "awaiting review" and act accordingly.

joshsmith:
I tried spending some time on this earlier today, but got really stuck. There are some issues I think with how we're approaching this now. Some things I'd like to do:
- Ensure uniqueness across `User` and `Organization` for the slugs
- [Make sure uniqueness does not create race conditions](http://stackoverflow.com/questions/34049308/how-to-avoid-a-race-condition-when-validating-uniqueness-across-two-tables-in-ra)
- Allow users to define their own slug for the organization, so I may want "codecorps" but not "code-corps"
- Have a different name for the organization than its slug, i.e. "Code Corps" and "code-corps"

I got stuck on multiple options for how to accomplish all this in the cleanest way possible. There are a bunch of major design decisions that need made here really. And everything I seemed to do was like hacks on hacks on hacks.

So here are my questions:
- Should our user/organization models be polymorphic? Seems like no, but I'm not sure.
- Should we create a `Slug` model like I laid out earlier in the SO question? And if we do, do we wrap all of our `create` and `update` in transactions? What's the best way to do this? Scenarios?
- Do we even use `friendly_id` here? Or just use some of their finder methods for ourselves if we're going to have users creating the slugs anyway? I don't see a lot of benefit to all the other things `friendly_id` provides, but the finders are quite nice since we _should_ have `GET /model/:id` routes in addition to the `GET /:slug` routes. If users are creating slugs (`username` is already slugged in the case of `User` and defining their own `slug` for `Organization` (or we just `parameterize` on front-end for them for a nice default)), then I'm not sure we get all the other benefits of `friendly_id`.

I know we don't get a lot of time to discuss this in depth, so I'm relying on you to make some decisions here @begedin and we can run with it. Or if you think there are other things to do in the meantime and want to wait, I'm okay with that, as well.

Whatever you decide, you have my support.

begedin:
@JoshSmith I actually quite like us having the parent `RouteName` model of some sort.

``` Ruby
create_table :route_name do |t|
  t.belongs_to :owner, polymorphic: true
  t.string :slug, null: false
end
```
- `owner` is either `User` or `Organization`
- Slug is set to `(owner.username || owner.name).parametrize` initially.
- Users/organizations can edit the slug on their own
- Not sure if `slug` should update automatically if `user.username` or `organization.name` updates, especially once explicitly edited already
- We have a `/:slug` route - it can even just return a `RouteName` instance, with `owner` included. Ember can support it as a polymorphic relationship to. 
  - If we just return the `RouteName` instance to the client, we can also expose `PATCH /route_name` for the purpose of editing the slug. No need for special behavior.
- We still maintain `users/:id` and `organizations/:id` routes
- No use for `friendly_id` anymore, really, unless we also want to maintain `users/:slug` and `organizations/:slug` routes.

joshsmith:
This seems like a good approach. I don't like the idea of Ember doing all the work since this especially should probably be a server-side job. What the `GET /:slug` feels like an implementation detail to me; we just expect it to return back the right resource, which doesn't seem like it should be an instance of `RouteName`, but the resource itself.

joshsmith:
We will indeed need to wrap creates and updates in transactions here, since the association is not present until the parent record is saved and available, in this case the `owner` of the `SlugRoute`.

joshsmith:
To do:
- [x] Create scenario for saving a `User`
- [x] Create scenario for saving an `Organization`

Scenarios should include creation/update of `SlugRoute`.
- [ ] Ensure scenarios are called from controllers
- [x] Write test to ensure a user can't be created with the same slug of an existing organization
- [x] Add routing
- [x] Fix routing tests to take into account new routes
- [x] Ensure users/organizations cannot pick names of designated present or future routes to avoid conflicts (like `/support`, for example)

joshsmith:
I can't make the routing tests work. Something leads to be believe since `call` is never hit that these tests don't really test that the routing works in full. It's doing some magic that doesn't work all the way through the dispatching. It'd be nice if we could get them to work since they're so exact in terms of what it seems like we need, so I marked them as pending. Would be good to see if you can figure it out @begedin.

joshsmith:
The routing that I did is far from ideal and feels a little hacky. A couple things I would like to see if we can improve:
- [x] Actually render out 404 JSON in place of the `raise ActiveRecord::RecordNotFound`

Also, I broke the `UserPolicy` and `OrganizationPolicy`, again because of something I couldn't figure out with how this was affecting routing magic. I had to explicitly call `authorize organization, :show?`, or else I would end up having the policy call `?` with no actual method name. You can try it for yourself and replicate, but it'd be nice to figure that out, too.
- [x] Figure out how to get policies to recognize the correct actions

begedin:
@JoshSmith The issue with most of these is that we're basically circumventing the routing layer with behavior of our own and calling the controller actions directly.
- We're not in a controller, so we can't `rescue_from ActiveRecord::RecordNotFound` in the application controller to render a 404
- We can't render it from our `SlugDispatcher` either, since we don't have access to response rendering methods from a controller there.
- Pundit is unable to infer which action needs authorization. I guess it happens in a step we're skipping when accessing the controller actions directly.

I like how `SlugDispatcher` is written and I think we definitely ought to keep it, but I think we need to somehow hook it into routing directly, so less of the normal routing process is circumvented. I'm looking into it at the moment.

begedin:
@JoshSmith 
- Initially, I was able to fix the issue with pundit, by modifying the request object further before dispatching and adding an action name value to it.
- I was unable to find a nice way to handle `RecordNotFound` which was raised within the dispatcher.
- I was unable to find a better way to write the `Render#call` method.
- I got rid of the routing spec and instead added a request spec for `/:slug` which asserts proper controller actions are being called based on context, since we're circumventing routing here.

At that point, I did a commit, but we're still not handling `RecordNotFound` here. Past this point, I tried something else.

I reverted to using a constraint based approach, except this time, I used some of your approach to write it in a bit less hacky way. We use a single constraint which tries to match to both routes. If it matches to Organization, the router maps to `organizations#show`, etc.

While matching, it also modifies the request parameters, adding an id with the value of `SlugRoute#owner_id` to it, meaning our organization and user controllers remain clean.

At the same time, if it doesn't match either, it will raise an `ActionController::RoutingError`, so I added a handler to our application controller for it, as well as support for it in our `ErrorSerializer`.

These changes are within a single, most recent commit, so if this approach is not acceptable, feel free to remove it.

joshsmith:
@begedin this is perfect. Really glad we solved this. I added back in the routing specs as well for just an additional layer of certainty here.

joshsmith:
I have decided to get rid of scenarios altogether for this, and instead wrap things in an `after_save` with a `.lock` on `SlugRoute`. This should roll back the records as expected.

New to-dos:
- [x] Write model specs for `User`
- [x] Write model specs for `Organization`
- [x] Add validation and `after_save` to `Organization`
- [x] Remove scenarios and scenario checks in tests
- [x] Write front-end tests for duplicate slugs

scarfdestroyer:
:+1:  :100:

Issue #16, opened 2015-07-14
Add GitHub repos to project endpoint


joshsmith:
We need to consume GitHub repos in Project information, at least on the individual project pages.

begedin:
**Blocked by #14**

PR #78, opened 2015-12-03
GitHub repository model

### LINKED ISSUE #14: Add GitHub repository model
We need a model that includes a `title`, `organization`, `repository name` for a GitHub repository. A project should be able to have many of these.

Not sure what the proper name for this is.

begedin:
Closes #14, #16 

I'm unsure about the actual fields a GitHub repository should have.

The way it is right now, there's a relationship to project, an `owner_name` (since owner can be a github user or an organization) and a `repository_name`. #14 suggests there should be a title, but a github repository does not have a title.
- [ ] Is the title something the user ought to enter themselves, sort of a descriptive name for the repo?

The `GET /projects/:id` now includes github repositories as part of the response. The repos are included both as relationships and as actual data, via `include: :github_repositories` in the `render` call.

There is a `POST /github_repositories` endpoint now, which behaves in the following way:
- It requires a project to be specified
- It requires a url to be specified
- Both of these requirements throw an `ActionController::ParameterMissing` error if they fail. *
  - The `url` is not an actual field, so any validation errors would involve `repository_name` and `owner_name`, which would be confusing to the user.
  - We need project information to determine if the user is allowed to add the repo in the first place, so the only other error we could throw for it is an `ACCESS_DENIED`, which would make no sense if no project is specified at all.
- It returns the newly created repo, without any includes.

I used the URL parsing code from our code gem (where I'm the one who wrote it, so I carry the blame for any bugs) and placed in in a helper in our /lib folder. My guess is, we'll eventually be using GithubAPI instead, which will allow us to fetch the information from the provided URL directly. At that point, we'll probably also want a url field to store it in the database.
- [ ] Our access right policy only requires the user to be the project owner. It doesn't take any team membership information into account, and will probably rely on contributor roles once #12 is closed anyway. **We may even consider this blocked until then.**


begedin:
@JoshSmith I also marked this as "awaiting review". If we want the title field, then it needs to be added to our endpoints and migration. If not, then this is ready for merge.

joshsmith:
@begedin I have some failing specs here I couldn't yet resolve. Most of them are due to n+1 queries on including project in the `ContributorsController`. Maybe we need a different serializer? I'll leave that up to you.

begedin:
To get rid of the N+1 query errors, I decided to not include the project or the user in a `POST /contributors` response.

My reasoning is, this endpoint is used to apply to a project as a user. In any scenario I could conceive, the client would already have the current user and the project in question fetched, so there should be no need to include them again.

Outside of that, there were some minor errors in various specs, which I fixed.

begedin:
@JoshSmith There are still some questions here:
1. Is the `title` a field we need to add? Something the user types in when addinng a repository to a project?
2. Should the endpoint be `POST /:member_slug/:project_slug/github_repositories` instead of just `POST /github_repositories`?
3. Who should be able to add github repositories to a project? Any contributor, or admin and higher? Right now, anyone that's admin or higher can do it.

joshsmith:
1. I don't think we need to add a `title`. Most people interact with github seeing `code-corps/code-corps-api` as a title already.
2. I think the endpoint is probably fine as-is? So long as we have the IDs we should be fine.
3. You got it right on permissions.

begedin:
@JoshSmith In that case, this ought to be mergeable.

I asked about the endpoint because I was under the impression the change would make it more consistent with other endpoints, but now I see the current form is more consistent than the alternative.

joshsmith:
:+1: :ship: :it: 

:shipit:

PR #87, opened 2015-12-05
Add sequencing to posts

### LINKED ISSUE #86: Add numbering to a project's posts
Posts in a project should be numbered according to the numbering scheme of published (or not drafted) posts in that project. That is, don't use the `id` for the post number, but use a separate number, ordered by `1,2,3`

joshsmith:
Closes #86 

This uses the `sequenced` gem to create a sequence of posts so we can use their numbering to make posts look a little nicer, and make their referencing a little easier.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,8 @@ gem 'koala'
 
 gem 'redcarpet'
 
+gem 'sequenced'
+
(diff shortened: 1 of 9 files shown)


scarfdestroyer:
:+1:

PR #90, opened 2015-12-05
Redo routing

joshsmith:
It's time to re-do routing, again! Because reasons.

We were going to have rewrite Rails routing with the approach we were taking before. The more nesting you do, the hairier it was going to get.

This introduces the concept of a `Member` (tell me you have a better name!) that's essentially just a parent object for either a `User` or an `Organization`, which is polymorphic with a `model`.

This will enable us to write URLs like `/:slug/:project_slug/posts/:sequential_id` which correctly preserves a `member_id` `project_id` and `id` parameter that maps to `slug` `project_slug` and `sequential_id` respectively. And we can nest as far as we want, by just using `find_by` with multiple attributes to ensure we don't accidentally retrieve a post for someone else's project, or a project owned by someone else.

There are plenty of things left to do to fill out this code spike, namely:
- [x] Remove `SlugRoute` model
- [x] Modify `SlugValidator` to accept a `Member`
- [x] Write model specs for `Member`
- [x] Write serializer specs for `Member`
- [x] Write API specs for `Member`
- [ ] Add proper 404s when a `project` is `nil` (or any other record, for that matter)
- [x] Add slugs to projects
- [ ] Rewrite all of our routes to nest properly
- [ ] Rewrite API specs to use proper nested routing
- [ ] Rewrite controller methods to `find_by` appropriately
- [x] Remove `SlugConstraint` model


begedin:
While I do agree with slight restructuring, do we really want to nest our routes this way? As far as I can tell, JSON API prefers flat routes, where each resource route is accessible from the root.

Nesting them on the API side means we will have to override `buildURL` in ember for each of our adapters. There's route nesting on the client side of things, but that's just presentation.

begedin:
@JoshSmith I'm a bit weary with keeping the "member" name to, especially since we already have `members` through `team_memberships` on `organization`, for instance. 

I also don't have a better name either. My first instinct would have me go with `entity`, but that's as generic as it gets.

begedin:
@JoshSmith I'm unclear on some of the things here.

Our route is `GET /:member_slug/:project_slug`. What about `PATCH` and `POST`? Do they remain
- `PATCH /projects/:id`
- `POST /projects/`
  or do we change them to 
- `PATCH /:member_slug/:project_slug` 
- `POST /:member_slug`,

or do we keep both? Based on `config/routes.rb`, it looks like we keep both?

There are also some additional uncertainties and changes.
- Project routes didn't require authentication at all. I added this requirement to POST and PATCH. Not sure if we need it for GET. My guess is, that one should be public, but then we pass the fetched project through authorization to determine if the project itself is accessible depending on various factors.
- When creating a project, we need to differentiate between creating it for a user or an organization. This is easily workable for a `POST /:member_id/` route, but not for as easy for a `POST /projects` route. With the first approach, we simply check what `member_id` points to, then handle authorization based on current_user and `member.model`. For the other route, we would have to specifically require at least an `organization_id`

Also, do we keep the `GET /projects` route? My guess is, we should have a `GET /:member_id` route, which returns the member and includes the projects with it.

begedin:
@JoshSmith I updated the serialization matcher so it works with pagination and meta content now.

At this point, I'm thinking it might be gem-worthy on its own.

I'm not sure where we're at with this otherwise, though. I'll scan through the code to see if I have any remarks and list them here.

begedin:
# Notes

These are just all of my observations listed in one place, so we know what we have implemented.
- `GET /:slug`
  - Does not require authentication
  - Returns the member, with associated model included
  - [x] Doesn't handle a 404 response at the moment, there is no spec for it. Added it
  - Had the `authorize` call commented out. I commented it back in, but it returns true for now
    - [ ] Figure out `MemberPolicy#show?`
  - **Should the included model include projects as well?**
- `GET /projects`
  - [ ] I'm thinking **this should go away completely?** Or do we wan't to use it in some way where a filter is required. I can't imagine a scenario where we fetch a list of all projects.
- We're missing a route that would fetch projects for an owner.
- `GET /:slug/:project_slug`
  - [x] Doesn't have a 404 response specs for a missing project or owner. Added those
  - Didn't call authorize in any of the actions. Added the calls, but we need to flesh out the logic
    - [ ] Figure out `ProjectPolicy#index?` if it actually remains a possible action
    - [ ] Figure out `ProjectPoicy#show?`
    - [ ] Figure out `ProjectPolicy#create?`
    - [ ] Figure out `ProjectPolicy#update?`
- `POST /projects`
  - [x] Added specs to ensure owner is specified in the request. This includes model specs and validation
  - [x] There was no 404 response in case owner doesn't exist. I refactored the controller code so `create_params` now provide an `owner` instead of `owner_id` and `owner_type`. This makes it so that a `RecordNotFound` will be raised when appropriate. Also added specs to cover this behavior.
- `PATCH /projects/:id`
  - [x] Was missing a spec to ensure a 404 response is returned when the project doesn't exist. Added it.
- `GET /projects/:id/posts` 
  - No authentication required
  - [x] Had a duplicate spec which I removed.
  - Responds with a 404 when project doesn't exist
  - Lists posts for that project
  - [ ] Need to flesh out `PostPolicy#index?`. Just returns true right now
- `GET /projects/:id/posts/:post_id`
  -  No authentication required
  - Returns that post, includes comments
  - [ ] Need to flesh out `PostPolicy#show?`. Just returns true right now.
  - Responds with a 404 if project doesn't exist
  - Responds with a 404 if the post doesn't exist
- `POST /posts`
  - Requires authentication
  - Requires specifying of `title`, `markdown` and `project`
  - Ignores `status` and sets it to open
  - Ignores and autosets `number`
  - Allows `post_type` but also defaults it to `issue` if not present
  - [ ] Need to flesh out `PostPolicy#create?`. Currently, it just requires user to be logged in.
  - [x] Cleaned up some of the specs a bit, to make use of new helpers we have.
- We do not have a `PATCH /posts/:id`. We likely should, but I'm thinking in a separate PR
- Not sure if we need a `DESTROY /posts/:id` or if we'll just be closing them. In we do, separate PR.

joshsmith:
On all these policies, shouldn't `show` generally be `true`? I mean, any user should be able to at least view these things, if not comment on them or do any real actions with them.

joshsmith:
I thing `GET /projects` stays as something that's filterable and pageable, so that needs to change. But we'll never get all of them at once.

begedin:
> On all these policies, shouldn't `show` generally be `true`? I mean, any user should be able to at least view these things, if not comment on them or do any real actions with them.

Generally yeah, but I'm not sure if things can be marked as "private" at any point. I'm considering a scenario where someone is trying to fetch a record that is marked as private and they don't have access to.

> I thing `GET /projects` stays as something that's filterable and pageable, so that needs to change. But we'll never get all of them at once.

That makes sense, although we don't have filtering yet. That's a good candidate for a separate issue, though.

joshsmith:
No, I think with these generally we want them to be viewable.

One exception maybe is the `GET` for a Post or Comment that's not published.

PR #91, opened 2015-12-07
add obscenity gem and test for non-obscene usernames

scarfdestroyer:


joshsmith:
I think this is missing specs for seeing what happens at the request level. I'm not confident I know. When I try to create a user with a profane username, what will I get back?

joshsmith:
:+1:

PR #93, opened 2015-12-08
Add migration, model and model specs for user relationships

### LINKED ISSUE #84: Add following/followers

begedin:
Closes #84 
# Implementation

Straightforward.

We have a `:user_relationships` table, with a `UserRelationship` model class. The model has `belongs_to` associations to users, one named `follower`, the other `following`.

Validations are in place
- `follower` and `following` are both required
- There can be no duplicate records (same `follower, following` pair.

Users also have new `has_many` relationships - `active_relationships`, `passive_relationships`, `followers` and `followed`.


joshsmith:
I changed `followed` to `following` since I think it's more common.

PR #101, opened 2015-12-08
Add comment user mentions

### LINKED ISSUE #100: Add comment user mentions
We need to do the same as in #73 but for comments.

joshsmith:
Closes #100 

We need to add:
- [x] Spec for GenerateUserMentionsForComment
- [x] GenerateUserMentionsForComment methods
- [x] Model specs for comment
- [x] Same methods in model as in Post
- [x] Model specs for comment user mention
- [x] Same methods in model as in PostUserMention
- [x] Serializer specs for comment serializer
- [x] Comment serializer
- [x] Request specs for comment
- [x] Ensure includes are working as expected

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -12,7 +12,7 @@ class PostsController < ApplicationController
     post = find_post!
     authorize post
 
-    render json: post, include: [:comments, :post_user_mentions]
+    render json: post, include: [:comments, :post_user_mentions, :comment_user_mentions]
   end
 
   def create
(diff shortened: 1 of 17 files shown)


joshsmith:
You'll notice I serialized the comments directly on the post.

I'm not sure we should be including all this crap in the `index`. Seems like overkill. But definitely on the non-index routes.

begedin:
:+1: Looks good to me

begedin:
Also, I agree with not including most of the stuff on the index. We should start splitting up our serializers.

joshsmith:
@begedin add an issue(s)?

begedin:
Created issue for serializer - #106

PR #102, opened 2015-12-08
Add notifications

joshsmith:
- [x] Add a polymorphic `Notification` model
- [x] Add model specs for `Notification`
- [x] Add a scenario for generating notifications for post user mentions
- [x] Add a scenario for generating notifications for comment user mentions (depends on #100)
- [x] Wrap scenario calls in workers
- [ ] Call workers in transitions on models (to published or edited states on saved)
- [ ] Write actual emails that will be sent
- [ ] Write tests for emails


begedin:
@JoshSmith I wrote the scenario for comment mentions, but I'm unable to test it, since the PR it depends on isn't merged yet. The specs have been written blindly, though.

I also wrote the workers which are supposed to wrap the scenario calls. I'm unable to run the workers from anywhere, though. They should run from the update controller actions, but that's also something we need to merge first.

Lastly, I wrote a very basic mailer and specs, just so we have a template. I believe we could write a single mailer for both types of notifications, but we may have to diverge at some point.

joshsmith:
@begedin #100 is now merged into this.

begedin:
Notes for myself, so I don't forget things implementing this
- [x] Ensure worker is called at appropriate times during post creation/edit
- [x] Ensure worker is called at appropriate times during comment creation/edit
- [x] Ensure post worker creates correct number of notifications, sends correct number of emails, dispatches notifications
- [x] Ensure comment worker creates correct number of notifications, sends correct number of emails, dispatches notifications

begedin:
@JoshSmith Some questions
- Do we use `.deliver_now` or `.deliver_later`. I'm unfamiliar with action mailer?
- \* I'm not sure the custom `Post#update!` method is appropriately named. It might be me, but it caused some confusion in my case. Maybe another name would be preferable?
- I'm not sure creating mentions should be tied to the save callback of a model. Stuff like that has already created problems with factories in the past. Maybe all behavior like that should be part of a worker/scenario? My guess is, it needs to be a scenario, or a synchronous worker, since we need to provide the data to the user immediately.
- **There is no `PATCH /comments/:id` nor is there state management for comments. This is something that should be done soon, but probably in a separate PR.** - #116
- **The email I have is very basic and not very useful.** It should probably contain more information and ideally, a link to the post or comment in question. I'm not sure how to generate the link, though. That could actually be a major task, since we would need some sort of link that works in various potential client apps. That alone is worthy of an issue by itself, but I'm not sure what our aim is for it.

joshsmith:
@begedin I agree that `update!` needs renamed. I'm not sure to what. Do you have an alternate idea?

I disagree about the save callbacks. I think it's more Rails-y and I went with the scenario approach first, but that caused its own set of problems. This makes more sense to me since it's all part of the same transaction and it should be failable if the mentions aren't generated correctly. I believe that if we're having problems with factory girl that's less a symptom of bad design in Rails and more a symptom of the way we're using factories. I have had no issues with factories here, though.

joshsmith:
I'm pretty sure I'll need to:
- [x] Rewrite PATCH specs to take advantage of `make_request` helper methods.
- [x] Ensure notifications are only run when publishing
- [x] Ensure notifications are run on both POST and PATCH
- [x] Ensure duplicate notifications are not created

joshsmith:
@begedin I made some minor changes here. Would be good if you looked over changeset and decide if this is good to go.

One thing left to do, though is change that damned `update!` method to something better.

joshsmith:
We do want to use `deliver_later`, but we need to tell Sidekiq to do so: https://github.com/mperham/sidekiq/wiki/Active-Job#action-mailer

joshsmith:
`PATCH comments/:id` is contained in here.

joshsmith:
I can work on the email later. For now it's good to merge and I'll improve.

URL generation, though, should be pretty straightforward. What does GitHub link to?

I'm guessing it'd just be a URL of the post with a `#comment-4` or something for the UI.

begedin:
@JoshSmith The problem with URL generation is that we don't have a site yet. We can give it a link to the endpoint which returns the post/comment as json, but there isn't anything else.

Once we do have an ember app up and running, for instance, we can use that. I believe mobile apps can handle a url to a specific domain as if it was a protocol, so that should not be a problem.

joshsmith:
Let's not worry about the mobile app link, but we should be able to figure out what the web URL is for now, and change it later if it does indeed change.

begedin:
@JoshSmith In that case, should figuring out the link be part of the future email PR, or be done here?

Outside of that, this looks ready to me :+1:

joshsmith:
Let's just do it as part of the email PR.

Issue #95, opened 2015-12-08
Write PostPolicy


joshsmith:
Specifically at least the `create`.

scarfdestroyer:
Potentially satisfied with pull https://github.com/code-corps/code-corps-api/pull/103

PR #103, opened 2015-12-08
add post policy and update posts spec and post controller to comply

scarfdestroyer:
From issue https://github.com/code-corps/code-corps-api/issues/95

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -16,8 +16,10 @@ class PostsController < ApplicationController
   end
 
   def create
-    authorize Post
     post = Post.new(create_params)
+
+    authorize post
+
     if post.save
       render json: post
     else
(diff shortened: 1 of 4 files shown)


begedin:
:+1: From me. Looks really good!

@JoshSmith Just to make sure, these are the current abilites defined by `PostPolicy`
- `index?`, `show?`
  - Anyone can do it. The methods return `true`
- `create?`
  - Any logged in user can do it for an `issue`. They don't have to be a contributor.
  - Any approved contributor (collaborator or higher level) can do it for other types of posts
- `update?`
  - Any user can update their own issue
  - Any admin or owner-level contributor can update other people's issues
- `destroy?`
  - Not defined at the moment, but we don't have a destroy action either, so no need.

If you feel these policies are good, I think this is mergeable.

joshsmith:
These are a-okay.

joshsmith:
:+1:

Issue #104, opened 2015-12-09
Refactor out sequenced


joshsmith:
We want to have the same `number` sequence for posts that currently exists, but only set it once it's `published`. This is blocked by #98.

See [this gem](https://github.com/djreimer/sequenced) for how they do it, and we can replicate from there.

joshsmith:
Actually, it looks like this can be done using a `skip` lambda and checking for the state in there.

PR #105, opened 2015-12-09
Add state and allow skipping numbers

### LINKED ISSUE #96: Add PATCH /posts/:id
This does not exist.

### LINKED ISSUE #98: Add a state machine to Posts
- [ ] Add a state machine to Post for draft and published
- [ ] Add some sort of edited state (don't think this belongs in state machine, but just maybe a bool for determining whether to show that a post has been edited)

joshsmith:
This will close #96, #98 and #104.

Instead of refactoring out `sequenced`, I opted to keep it and instead use the `skip lambda`, which requires doing some things with the indexes (unfortunate but necessary) and adding a new validation. This should ideally prevent collisions. I think postgres' locking will also prevent it, which is implemented in the gem.

Also adds a `published` and `draft` state.

We should still:
- [x] Fix couple of tests I lazily left failing
- [x] Add an `edited` bool, or make it a state if we can scope published posts to both published/edited
- [x] Add an `edited_at` timestamp whenever an `edited` post is saved
- [x] Create `PATCH` method to update the published/edited states
- [x] Write serializer specs to ensure the state is getting passed on. This shouldn't be named `aasm_state`, but wrapped somewhere (model?) to not be exposed directly
- [x] Write API-level specs to ensure we're setting this information correctly

And anything else I'm missing here.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -1,6 +1,6 @@
 class PostsController < ApplicationController
 
-  before_action :doorkeeper_authorize!, only: [:create]
+  before_action :doorkeeper_authorize!, only: [:create, :update]
 
   def index
     authorize Post
@@ -25,7 +25,26 @@ class PostsController < ApplicationController
     end
   end
 
+  def update
+    post = Post.find(params[:id])
+    authorize post
+
+    post.assign_attributes(update_params)
+
+    if post.valid?
+      post.edit
+      post.save!
+      render json: post
+    else
+      render_validation_errors post.errors
+    end
+  end
+
   private
+
+    def update_params
+      record_attributes.permit(:markdown, :title)
+    end
 
     def create_params
       record_attributes.permit(:markdown, :title, :post_type).merge(relationships)
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -41,11 +41,11 @@ class PostsController < ApplicationController
   private
 
     def update_params
-      record_attributes.permit(:markdown, :title)
+      record_attributes.permit(:markdown, :title, :state)
     end
 
     def create_params
-      record_attributes.permit(:markdown, :title, :post_type).merge(relationships)
+      record_attributes.permit(:markdown, :title, :state, :post_type).merge(relationships)
     end
 
     def project_relationship_id
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -31,9 +31,7 @@ class PostsController < ApplicationController
 
     post.assign_attributes(update_params)
 
-    if post.valid?
-      post.edit
-      post.save!
+    if post.update!
       render json: post
     else
       render_validation_errors post.errors
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,8 @@ gem 'obscenity'
 
 gem 'sequenced'
 
+gem 'aasm'
+
 group :development, :test do
   gem 'dotenv-rails'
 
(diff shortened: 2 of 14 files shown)


joshsmith:
We should probably try to close #96 here.

begedin:
@JoshSmith Since our post has timestamps, could we not simply use that to determine the post was edited? Basically, 

``` Ruby
def edited?
  updated_at != created_at
end
```

My guess is, we need it to show that type of stratus on the client and for not much else. We could even define a custom scope using that. It should not be too costly, in my opinion.

joshsmith:
I don't think that's sufficient. It's `edited` in the sense that since it's been published, it has been edited (so that we can show this on the front-end). If you go and edit your comment since I reply to it, it'd be nice to know this was the case.

So this needs done not just on this model, but `Comment` as well.

joshsmith:
We also want to see the time at which it was `edited`, which is `nil` in some cases and `updated_at` I suppose if it was actually `edited`.

begedin:
@JoshSmith Yeah, I've jumped to that a bit too quickly there.

One option (without the column) is

``` Ruby
event :publish do
   transitions from: :draft, to: :published 
end

event :edit do
  transitions from :published, to: :edited
end

# and then
scope :active { published.merge edited }

# This would also give us the option to do

def edited_at
  self.updated_at if self.edited?
end
```

The other option, with the column would be

``` Ruby
scope :edited { where(edited: true) }
scope :active { published.merge edited }
```

I don't think one is much faster than the other, so I'm sort of leaning towards the one without the extra columns, since all other things seem to be equal.

Depending on how we want to display the post/comment, we may need a `published_at` stamp as well, which can be set during the `:publish` event. That would probably be an extra column. I'm thinking we could use it to display the date the post was published at. On the other hand, if we want to display just the date when the post was last edited, then using the `updated_at` column should be sufficient.

begedin:
I added the stuff listed here, using the following approach
- `edited` is an `aasm` state. Only published posts can become edited.
- `edited_at` is a method on the post model. It returns `updated_at` if the post is in the edited state, nil otherwise
- `state` is a post method which wraps `aasm_state`
- `PostSerializer` and `PostSerializerWithoutIncludes` now also have the `state` and `edited_at` properties.
- In order to publish a post, it needs to be in `edited` state and we need to `PATCH /posts/:id` with a payload containing `state: "published"` as a parameter. I'm not liking this, but I'm not sure what a better approach would be? A specific API endpoint we could hit with a `post_id` just for the purpose of publishing?

If we want to use an edited column approach, then we just remove the state, modify the `active` scope and add the migration. I believe `ActiveRecord` auto-generates an `edited?` method if there's a boolean column named `edited`.

joshsmith:
@begedin Won't `update` be called on posts when in a draft mode? In which case `post.edit` will error out in those cases?

app/controllers/posts_controller.rb
```diff
@@ -25,8 +25,27 @@ def create
...
+    post.assign_attributes(update_params)
+
+    if post.valid?
+      post.edit
+      post.save!
+      render json: post
+    else
+      render_validation_errors post.errors
+    end
+  end
```
joshsmith, line 39: I think you can just call `if post.edit!` here, but doesn't address the issue I had involving `edit` where posts can sometimes still be in a `draft` state, i.e. every time you click "Preview" on the client side.

joshsmith, line 39: Maybe we create an `update!` method that wraps these two different behaviors depending on what the state was.

joshsmith:
I've added an `update!` method and a writer for `state`, and now need to add tests in controller with `permit`.

scarfdestroyer:
:raised_hands: :hankey: :+1:

PR #113, opened 2015-12-09
Add state machine to comments

### LINKED ISSUE #111: Add state machine to comments
We need to add the same sort of state machine that posts have to comments.

joshsmith:
Closes #111 
- [x] Model spec for comment
- [x] Serializer spec
- [x] GET specs
- [x] PATCH specs
- [x] update method


joshsmith:
This really needs a thorough review as I just powered through it.

joshsmith:
Honestly these specs probably could use some rethinking/refactoring, too. They're getting verbose.

begedin:
Seems good to me. :+1: 

Just keep in mind, if we merge this first, then #102 will have to be modified to accommodate for it. If we go with #102 first, then we'll have to update this one before merging. There's overlap.

Issue #94, opened 2015-12-08
Write ProjectPolicy


joshsmith:
We need to do the `index`, `show`, `create`, and `update`.

begedin:
## @JoshSmith I wrote a proposal in the associated PR #114, but here it is, copied in full

## Two general cases

### User is project owner
- The owner needs to be the current user.

### Team is project owner
- The current user needs to belong to the team and be able to define projects for that team. This should depend on their role in the team, so `user.team_memberships.find_by!(team).role`. The problem is, there is currently no such thing as a membership role in our `team_memberships` table.
- If the current user does not belong to the team, it's possible they could belong to the organization who owns the team, and they could have a role in the organization which would still allow them to define projects for that team. Again, this isn't something we have the database for right now. The `organization_memberships` table would require a column for it.

This would mean we need to 
- [ ] Define possible role levels for  `team_memberships` table and add a column to it
- [ ] Define possible role levels for `organization_memberships` table and add a column to it
- [ ] Decide if there's an organization role level which allows the user to create projects for a team they aren't a member off.

PR #114, opened 2015-12-10
Add project policy

scarfdestroyer:


scarfdestroyer:
Potentially satisfying issue https://github.com/code-corps/code-corps-api/issues/94

begedin:
@scarfdestroyer Not sure how familiar you are with GitHub, but if you write down something like `Closes #94` at the top of your PR description, GitHub will detect this and auto-close the issue with that number when you merge the PR.

https://help.github.com/articles/closing-issues-via-commit-messages/

I know the title says "via commit messages" but it works with PR descriptions to.

In any case, it helps to reference the issues a PR relates to.

begedin:
I don't think this is the policy we need. That being said, I'm also not completely sure what it should be.

`user.admin?` is not something we generally look at for policies. Admins are just special types of users that do special stuff in the system.

In this case, I believe we should have two general cases.
- If the project.owner is a user, then it needs to be the current user, because a user can define projects of their own.
- If the project owner is a team, then 
  - The user needs to belong to the team and be able to define projects for that team. This should depend on their role in the team, so `user.team_memberships.find_by!(team).role`. The problem is, there is currently no such thing as a membership role in our `team_memberships` table.
  - If the user does not belong to the team, it's possible they could belong to the organization who owns the team, and they could have a role in the organization which would still allow them to define projects for that team. Again, this isn't something we have the database for right now. The `organization_memberships` table would require a column for it.

joshsmith:
I think we need `user.admin?` here since we're restricting creation of projects for now.

I'm having trouble wrapping my head around just about everything else here.

begedin:
I was not aware of that. My impression was that site admins are needed to create organizations, not projects. Shouldn't any user be able to create a project for themselves, same as a GitHub user would be able to, on github?

joshsmith:
But for now, we're going to be limiting it. Eventually yes. Same with organizations. Eventually people will be able to create organizations freely. For now, though, it will be very closed.

begedin:
In that case, based on the current PR state, we have this set of rules
- If project owner is user, it has to be the logged in user
- If project owner is an organization, the user creating it needs to be an admin

In hindsight, that makes a lot of sense to me.

If that's the set of rules we want, then this feels ready for merge to me :+1:

joshsmith:
I do think, though, there's a gap like you said: what about teams of admins for organizations?

begedin:
For that part, we would need to expand the membership models (`team_memberships`, `organization_memberships`). At the moment, all memberships are made equal, since there is nothing to indicate a role.

joshsmith:
Ah, so there's no concept of a team being `admin` or `collaborator` or something? I guess we'll need to do that. Maybe write a separate issue for that, and one to modify the Project policy once it's unblocked by that.

begedin:
Do we merge this, then? Without memberships having roles, this is a good set of rules to begin with, in my opinion.

joshsmith:
Agreed. :shipit: And then just create those issues when you have :hourglass_flowing_sand:

begedin:
Created issues #119, #120. Merging this now...

PR #115, opened 2015-12-10
Add images for posts and comments

### LINKED ISSUE #74: Add images for posts and comments
A `PostImage` or `CommentImage` should be able to be uploaded as a Base64 encoded file. We'll take the original filename and save that.

We'll be replacing the first occurrence of a string like `` in the client-side with the actual URL of the image. I tested this on Github and if you paste that string and then drop the file in (which generates that string again), it will replace just the first instance of it, every time.

We can create a reference to a post or comment with this model as `belongs_to`, but it will never be used in a `has_many`. We should also probably make it `belongs_to` a `User`.

The image processing will need to be asynchronous as usual, but we'll have to subscribe to some async stream for it. I think some minimal resizing down should probably be done.

The URL string should absolutely have the ID of the post or comment and the ID of the image. If we ever change URLs, we may have to rely on this (with some downtime) to do URL replacement in our `markdown` and `body` text.

But we should be trying to make these URLs as immutable as possible.

joshsmith:
Closes #74 

I can't seem to get this working correctly. It looks like there's a Paperclip way to handle base64 strings by default. We want to keep the filename to be able to use that information for replacing the string, as it says in #74. And we also want to keep the original file type, not just turn it into a PNG.


begedin:
@JoshSmith 

I looked into support for base64 data with paperclip and it turns out, there is undocumented support, but only if the base64 string is in the standard data-uri format for images, meaning something like:

```
"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAfQAAAH0CAYAAADL1t+
KAAAKQWlDQ1BJQ0Mg\nUHJvZmlsZQAASA2dlndUU9kWh8+9N73QEiIgJfQaegkg0j
tIFQRRiUmAUAKG\nhCZ2RAVGFBEpVmRUwAFHhyJjRRQLg4Ji1wnyEFDGwVFERe
XdjGsJ7601896a\n/cdZ39nnt9fZZ+9917oAUPyCBMJ0WAGANKFYFO7rwVwSE8vE9
wIYEAEO..."
```

In that case, paperclip figures out the default adapter. Besides that, I don't think there's a way to embed the filename into the base64 data, so we need the original filename as a separate request parameter, but it seems you already accounted for that.

I modified `decode_image_data` so it should work by
- storing the original filename, with extension
- storing the image data

However, having the parameter contain the initial `"data:image/png;base64,"` part doesn't give much benefit API-side. I guess the client could be dumber in that case and just pass along the string, but depending on implementation, it might actually create a requirement on the client to add the string in the first place. 

Really, the only difference between our usual approach and the current one is that our `decode_image_data` is slightly simpler.

``` Ruby
def decode_image_data
  return unless base_64_photo_data.present?
  data = Paperclip.io_adapters.for(base_64_photo_data)
  data.original_filename = self.filename
  self.image = data
end
```

As you can see, we still need to set the original filename.

I'm unable to write proper specs for it or test further, since I don't have the proper AWS credentials.

There's also the additional problem of something triggering a `HEAD` request to the resulting image url. I added VCR cassettes to the two specs that trigger it to get around it, but since I don't have the proper credentials, the returning values are "403 forbidden". Since we already mock our `Attachment#save` calls to just return `true`, I'm not sure why it's happening in the first place. Really, it should not be happening at all.

joshsmith:
I still need to:
- [x] Add `CommentImage` model and spec
- [x] Add comment image factory
- [x] Add notify pusher of comment image scenario and worker and specs
- [x] Add comment image controller and API specs
- [ ] Add an attachment validator: validates_with AttachmentSizeValidator, attributes: :photo, less_than: 10.megabytes

joshsmith:
I'm not sure I like the way I'm testing the scenario getting called from the image worker.

JoshSmith:
Also would like to change the way I'm notifying pusher only once the image is decoded. That seems really tightly coupled.

joshsmith:
The attachment validator can come in a separate PR. This is already ginormous.

joshsmith:
Attachment validator is listed in #132 and policies in #131 #129

joshsmith:
Otherwise this is good to review, and needs a really thorough look for anything I might have missed, or written poorly.

spec/factories/comment_image_factory.rb
```diff
@@ -0,0 +1,17 @@
+FactoryGirl.define do
+
+  factory :comment_image do
+    association :user
+    association :comment
+  end
+
+  trait :comment_with_s3_image do
```
begedin: I'd rename this to `with_s3_image`, so it's consistent with `post_image_factory`.

begedin:
Other than that one potential rename, I didn't notice any issues. Considering the size of the changeset, though, I can't guarantee it. Cautious :+1: from me.

spec/factories/comment_image_factory.rb
joshsmith: For some reason the trait won't work with the other trait that exists. Think it's because it's not wrapped in the factory itself? I don't know.

PR #122, opened 2015-12-11
Added users to post

### LINKED ISSUE #53: Post should include a project's users
We'd like to add the ability to @mention at some point. This will require an `includes` for a project's user for any `Post`, since we need to be able to quickly reference a list of associated users for the project.

begedin:
Closes #53 

It's pretty straightforward, but I'm getting a bit worried about the amount of relationships we're starting to have.

The relationship chain here is actually \* `:post` -> `:project` -> `:contributors` -> `.map_by(&:user)`

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -12,6 +12,8 @@ class Post < ActiveRecord::Base
   has_many :post_user_mentions
   has_many :comment_user_mentions
 
+  has_many :users, through: :project
+
   acts_as_sequenced scope: :project_id, column: :number, skip: lambda { |r| r.draft? }
 
   validates_presence_of :project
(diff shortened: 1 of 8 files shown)


app/models/project.rb
```diff
@@ -4,6 +4,7 @@ class Project < ActiveRecord::Base
   has_many :posts
   has_many :github_repositories
   has_many :contributors
+  has_many :users, through: :contributors
```
begedin, line 7: In my opinion, it makes more sense to have users on projects and then get them to the post through the project, then to have them on the post directly. We will likely use them here for something in the future anyway.

app/models/post.rb
```diff
@@ -12,6 +12,8 @@ class Post < ActiveRecord::Base
   has_many :post_user_mentions
   has_many :comment_user_mentions
 
+  has_many :users, through: :project
```
begedin, line 15: See comment bellow. It makes more sense to me to get them to the project.

An alternative would be to get them here directly, but that would mean

``` Ruby

belongs_to :project
has_many :contributors, through: :project
has_many :users, through: :contributors
```

PR #124, opened 2015-12-11
Comment policy

### LINKED ISSUE #112: Add CommentPolicy
We need a policy for comments.

begedin:
Closes #112 
# Implemented policy
- Anyone can `index`
- Anyone can `show`
- Any logged in user can `create`
- Only comment authors can `update` their own comments

### DIFF
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -6,7 +6,21 @@ class CommentPolicy
     @comment = comment
   end
 
+  def index?
+    true
+  end
+
+  def show?
+    true
+  end
+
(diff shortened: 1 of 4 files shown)


app/policies/comment_policy.rb
```diff
@@ -6,7 +6,21 @@ def initialize(user, comment)
...
+    true
+  end
+
+  def show?
+    true
+  end
+
   def create?
     user.present?
   end
```
joshsmith, line 19: I believe this should be the same as `update?`, otherwise you can create a comment for another user.

PR #135, opened 2015-12-13
Add ember deploy

joshsmith:
This adds the ability to serve up our index file for the ember app from Redis using our API. If the app is deployed, then it renders the latest. This works in development, as well.

We still need to:
- [x] Write a request spec that tests the not found case
- [x] Write a spec that tests the development case
- [x] Write a spec that tests the revision case
- [x] Write a spec that tests the current case

Typically the keys in `redis-cli` look like:

```
127.0.0.1:6379> KEYS code-corps*
1) "code-corps-ember:index:revisions"
2) "code-corps-ember:index:current"
3) "code-corps-ember:index:__development__"
```


joshsmith:
I'm also not sure about the way I'm doing the `render`. I changed to `text` and set the response format in the `EmberIndexController`. I think this needs to happen to override the `json` formatting.

In some edge cases before with `lvh.me` I was getting text rendered and a `Content-Type` did not appear to be set in the response. We definitely need to check `Content-Type` in our specs.

begedin:
I added a spec. I'm mocking Redis manually instead of installing something like `MockRedis`, since the mocks consist of a single message.

For each of the four cases, I test that the correct redis key is being fetched, the correct response status and content type are set and that the response body has the mocked content I provided (simple string).

joshsmith:
@begedin does the way I'm doing the renders look right to you? For some reason in the Yapp talk about this they mentioned rendering as `text` with a format of HTML, but couldn't seem to make that work. I guess this is working well, though.

PR #138, opened 2015-12-14
Post comment image policies

### LINKED ISSUE #129: Add PostImagePolicy
We need a policy for images. We also need to uncomment the `authorize post_image` in the controllers.

begedin:
Closes #129, #131

Straightforward enough. 

User can add image to post or comment if they are the owner of that post or comment. Would that be about right, or do I need to modify it further?


joshsmith:
This looks good to me :+1:

PR #142, opened 2015-12-15
Made posts_controller#index use the `PostSerializerWithoutIncludes`

### LINKED ISSUE #106: Serialize posts_controller#index response with a custom serializer which does not contain unnecessary includes
Not actually sure if we should go with a custom serializer, or just use the same one, but not include the usual includes. We should probably research what the convention is.

The way I see it, with a has_many relationship, we would need big database fetches, which isn't great, so  our serializer probably shouldn't have any `has_many` relationships. `belongs_to` should be fine, since the id for that is contained within the record.

Maybe these lower-footprint serializers for index methods should only contain `belongs_to` relationships, no `includes` and `x_counter` fields for their `has_many` relationships?

begedin:
Closes #106 
# Implementation

We already had the serializer and its spec implemented for the purposes of using it in a relationship with some other object (likely project). However, we did not use it in `posts#index` yet.

We also had a bug in the serializer matcher. Turns out, even though the parameter to specify a serializer when rendering a collection is `each_serializer`, when initializing it directly, it was just `serializer`. We used `each_serializer`, which was getting ignored, so the default serializer was just being used instead. For all our existing specs, this was not a problem, but it became one now, so I had to resolve it.


begedin:
This should now be mergeable. The only hound errors are those indicating long lines of code, which is something I don't think we should too strictly adhere to. Making the errors go away here would involve writing strings that span across multiple lines.

begedin:
@JoshSmith This one should be good to merge

PR #143, opened 2015-12-17
Specs for cors

### LINKED ISSUE #133: Find a way to test rack-cors
I found something on the internet but it conflicts with our helper methods:

```
require 'rails_helper'

describe "the requests support CORS headers", type: :controller do
  scenario 'Returns the response CORS headers' do
    get '/people/1234', nil, 'HTTP_ORIGIN' => '*'

    expect(last_response.headers['Access-Control-Allow-Origin']).to eq('*')
  end
end

describe "the requests support CORS headers", type: :controller do
  scenario 'Send the CORS preflight OPTIONS request' do
    options '/', nil, 'HTTP_ORIGIN' => '*', 'HTTP_ACCESS_CONTROL_REQUEST_METHOD' => 'GET', 'HTTP_ACCESS_CONTROL_REQUEST_HEADERS' => 'test'

    expect(last_response.headers['Access-Control-Allow-Origin']).to eq('*')
    expect(last_response.headers['Access-Control-Allow-Methods']).to eq('GET, POST, PATCH, OPTIONS')
    expect(last_response.headers['Access-Control-Allow-Headers']).to eq('test')
    expect(last_response.headers).to have_key('Access-Control-Max-Age')
  end
end
```

begedin:
Closes #133 when done

This is an example of how we could test cors, but I need to actually write all of the tests before we can merge.

I'm not sure where we should put these, though. Should they go in a single feature context at the top level of that API spec (how it is right now), in a feature context for each of the individual method contexts, or in a separate file altogether?


begedin:
@JoshSmith I made an attempt and resolving rubocop warnings made by houndci. I guess the configuration houndci uses doesn't match the default one, so I'll delay doing this until I get the proper one.

begedin:
@JoshSmith I made an attempt to work out how hound works here and added a custom config for rubocop, which hound will read, since of the rules it enforces by default are, in my opinion. slightly too different from our usual style.

For instance, the comments that haven't been made oudtated above want us to use a trailing dot when chaining methods, which I don't think we've ever done:

``` Ruby
expect(something).to be_something.
  with_some_thing_1.
  with_some_other_thing_2.
  with_the_last_thing
```

Instead, we always did 

``` Ruby
expect(object).to be_something
  .with_some_thing_1
  .with_some_final_thing
```

I've added a few other rules to .rubocop.yml, which are explained in the file.

If you disagree with any of these, please let me know.

Outside of that, I'm still unsure how you want this to be tested. I could even go as far as making a simple matcher which would allow us to write something like:

``` Ruby
expect(route_path).to be_cors_compatible.for_actions(:get, :patch, :post)
```

begedin:
This should be mergeable

PR #153, opened 2016-01-01
Skip relevant tests if the missing the config

justinabrahms:
This allows people to have passing tests after first installation.

### DIFF
--- a/spec/models/comment_image_spec.rb
+++ b/spec/models/comment_image_spec.rb
@@ -1,4 +1,5 @@
 require "rails_helper"
+require_relative "../utils"
 
 RSpec.describe CommentImage, type: :model do
   let(:gif_string) {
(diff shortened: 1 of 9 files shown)


justinabrahms:
It'd also be really rad if I could run this linting locally, because getting 20+ emails on PR opening is frustrating.

joshsmith:
@justinabrahms I agree. I think as nice as Hound is for helping to enforce some conventions, it's kind of a PITA. @begedin has been grumbling about it for awhile. I think we just need to pull it altogether.

begedin:
We could still be using hound, but we should either set it up to user our own config, or we should switch to using hound's config.

I'm not sure we gain much from it, though. Local linting should suffice, if we take it seriously enough.

In any case, we need to agree on a common set of styles.

justinabrahms:
Simply having something I could run locally would be good for what I
wanted.

On Monday, January 11, 2016, Nikola Begedin notifications@github.com
wrote:

> We could still be using hound, but we should either set it up to user our
> own config, or we should switch to using hound's config.
> 
> I'm not sure we gain much from it, though. Local linting should suffice,
> if we take it seriously enough.
> 
> In any case, we need to agree on a common set of styles.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/code-corps/code-corps-api/pull/153#issuecomment-170451288
> .

## 

Sent via mobile phone.

joshsmith:
After some discussion, we've opted to close this in favor of finding ways to remove ENV reliance.

PR #159, opened 2016-01-04
change production.rb to force ssl

scarfdestroyer:
@JoshSmith I think this is what we need, take a look to make sure


config/environments/production.rb
```diff
@@ -42,7 +42,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = (ENV['FORCE_SSL'] == 'true')
```
joshsmith, line 45: @scarfdestroyer I'd make these suggested changes. Otherwise :+1:

joshsmith:
After changes, merge, open PR on master, and merge that when ready to get it up on production.

PR #162, opened 2016-01-11
Add config initializer for json api mime type

### LINKED ISSUE #161: Add mime type config initializer for json api
ember-data is pretty much the only API accessiing library I've used up to this point which actually sends post and patch requests with the content type header set to the proper `'application/vnd.api+json'` value.

AMS, however, ignores such a content type by default, so we need to add support for it via an initializer.

begedin:
Resolves #161


joshsmith:
:+1:

PR #188, opened 2016-01-19
Enable filtering by type alongside paging for posts endpoint

### LINKED ISSUE #196: API should only show published posts on post#index

begedin:
Closes #196 

This PR goes alongside code-corps/code-corps-ember#42

It enables filtering by post type while still supporting paging on the `GET /project/posts endpoint`


begedin:
@JoshSmith I would wait before merging this one until the associated ember PR is merged. It might turn out there are some other minor changes needed.

There's also the option of merging this and opening a new PR if that happens, though.

begedin:
There was an issue in how our `meta_for` method in `application_controller` was working. Instead of returning the total number of filtered records, it was returning the total number of all records.

The only solution I could think off was to pass in the count of unfiltered records manually, since that would depend on the controller calling the method, as well as the type of filter.

joshsmith:
:+1:

PR #189, opened 2016-01-19
Seeds for a basic organization, project and project posts

begedin:
Just some basic seeds we can use in development.

Added:
- a non-admin user
- a single organization
- a single project for that organization
- 5 of each types of posts for the created user and project


begedin:
@JoshSmith As far as I'm concerned, this is mergeable. The seeds will evolve over time anyway, and the ones defined here serve their purpose for now.

PR #192, opened 2016-01-19
Add ability to update user photo and associated specs

### LINKED ISSUE #187: Allow user to edit their photo
Currently, the API only supports `base64_photo_data` when creating a new user. It should be allowed when updating an existing user as well.

scarfdestroyer:
Changed `photo_param?` method. I don't think this would cause any problems but could add a `update_photo_param?` method alternatively.

Also, not sure if we need a spec for the profile picture update worker when we update a `user` without a photo.

Closes #187


spec/requests/api/users_spec.rb
```diff
@@ -405,10 +405,12 @@
 
     before do
       @current_user = create(:user, email: "current@mail.com", password: "password", website: "initial.com", biography: "Initial", twitter: "@user")
+      file = File.open("#{Rails.root}/spec/sample_data/default-avatar.png", 'r')
+      @base64_image = Base64.encode64(open(file) { |io| io.read })
```
scarfdestroyer: @JoshSmith You know how to fix this? These lines are present in a lot of the specs.

joshsmith: `.open(file, &:read)`

joshsmith: `Base64.encode64(open(file, &:read))`

joshsmith:
:+1:

joshsmith:
Make sure to say that this closes #187 in the description.

PR #198, opened 2016-02-01
Fix routing

### LINKED ISSUE #197: Routing needs to be nested
The Ember app has URLs like `code_corps/code_corps/posts/18`, which we can break this down into its component parts: `:slugged_route_slug/:project_slug/posts/:number`.

When making a request to the API, the Ember app, if hit directly at that URL, has _no knowledge_ of the `Project` `id`, or of any `id` at all!

This means that in order to find the project for a particular post, you would first have to do something like the following:

```
      Organization.find_by_slug(slugged_route_slug)
        .projects
        .find_by_slug(project_slug)
```

This means we need to change our routes to something like the following:

```
    resources :slugged_routes, path: '', only: [:show] do
      get "projects", to: "projects#index"
      resources :projects, path: '', only: [:show] do
        resources :posts, only: [:index, :show]
      end
    end
```

I don't see any way around this given the constraints we have.

We need to make a listing of all the possible routes, and what the Ember app could possible know or not know.

joshsmith:
This attempts to fix #197.

Far from finished, but spiking to see what routing looks like.


joshsmith:
Not too bad. Only failures are on API `projects_spec`, `posts_spec`, and `comments_spec`. If we fix those, I think we're in business.

joshsmith:
Closing because fixes in Ember allow us to keep route structure as is.

PR #202, opened 2016-02-03
Fix bugs

### LINKED ISSUE #200: Test comments controller properly
For some reason we have an `authorize Comment` that's not being hit in the `index` of the comments controller. This needs fixed.

joshsmith:
This fixes #200 and fixes an issue with `number` not being included in one of the post serializers.


scarfdestroyer:
Looks good, is there a reason the post is created with an explicit id?

joshsmith:
I think just to make it explicit. You'd have to ask @begedin since he did that originally.

scarfdestroyer:
:+1:

PR #205, opened 2016-02-04
Fix post numbering

### LINKED ISSUE #201: Sort posts by number descending
We need to write a test for this, and then ensure that posts are sorted by `number` desc.

joshsmith:
This fixes #201.


scarfdestroyer:
I'm not super knowledgeable with factories but looks good :+1:

PR #210, opened 2016-02-08
Fixes image decoding for projects and users

### LINKED ISSUE #209: Rewrite Paperclip data URI processing
We need to attempt something like:

``` ruby
  # Instantiates Paperclip::DataUriAdapter attachment
  file = Paperclip.io_adapters.for(data_uri)
  file.original_filename = name

  # Save the model  
  model.attachment = file
  model.save!
```

We need to test it with fixtures that have the `data` URI and fixtures without it. I believe iOS clients may send a base64 string that does not include the URI.

joshsmith:
This fixes #209.

General idea is that we may take in base64 data that does or does not have a URI in it, and need to decode that data accordingly.

I don't like manually specifying PNG but I don't see a way around that for right now.


rjsamson:
:+1: :+1: :+1:

Issue #213, opened 2016-02-08
Error during db:seed_fu script


ptrikutam:
Looks like the `posts.rb` fixture is executed before the `users.rb` fixture populated, causing the error below.

```
$ bundle exec rake db:seed_fu

== Seed from /Users/pavan/Development/code-corps-api/db/fixtures/organizations.rb
 - Organization {:id=>1, :name=>"Code Corps", :slug=>"code_corps"}

== Seed from /Users/pavan/Development/code-corps-api/db/fixtures/posts.rb
 - Post {:id=>0, :title=>"A task", :body=>"This is a basic task", :markdown=>"This is a basic task", :project_id=>1, :user_id=>2, :post_type=>"task"}
rake aborted!
ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  insert or update on table "posts" violates foreign key constraint "fk_rails_5b5ddfd518"
DETAIL:  Key (user_id)=(2) is not present in table "users".
: INSERT INTO "posts" ("aasm_state", "id", "title", "body", "markdown", "project_id", "user_id", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING "id"
/Users/pavan/.rvm/gems/ruby-2.2.3@codecorps/gems/seed-fu-2.3.5/lib/seed-fu/runner.rb:46:in `eval'
/Users/pavan/.rvm/gems/ruby-2.2.3@codecorps/gems/activerecord-4.2.5/lib/active_record/connection_adapters/postgresql_adapter.rb:602:in `block in exec_cache'
/Users/pavan/.rvm/gems/ruby-2.2.3@codecorps/gems/activerecord-4.2.5/lib/active_record/connection_adapters/abstract_adapter.rb:472:in `block in log'
/Users/pavan/.rvm/gems/ruby-2.2.3@codecorps/gems/activesupport-4.2.5/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
/Users/pavan/.rvm/gems/ruby-2.2.3@codecorps/gems/activerecord-4.2.5/lib/active_record/connection_adapters/abstract_adapter.rb:466:in `log'
/Users/pavan/.rvm/gems/ruby-2.2.3@codecorps/gems/activerecord-4.2.5/lib/active_record/connection_adapters/postgresql_adapter.rb:601:in `exec_cache'
[...]

PR #214, opened 2016-02-08
Explicitly order database fixtures

ptrikutam:
Fix for #213


joshsmith:
There's a further bug in posts.rb that starts numbering from 0, so that needs fixed.

And we should also be setting post.aasm_state = "published" in these fixtures, as well.

ptrikutam:
@JoshSmith -- removed the explicit setting of IDs in the posts.rb entirely & added the published status to all the posts.

joshsmith:
:+1:

PR #216, opened 2016-02-08
Update bin/setup script.

ptrikutam:
This converts the `bin/setup` script to a bash script. It also:
- Checks for the various dependencies required to run the server
- Drops & re-create the database
- Seeds the database

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,14 +22,12 @@ The Code Corps API is an open source Rails::API backend that powers the Code Cor
 
 ### Before you start Rails
 
-1. `bundle install`
-2. `bundle exec rake db:create db:migrate db:test:prepare db:seed`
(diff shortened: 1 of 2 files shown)


joshsmith:
Do you know if there's any real way to test this? I'm not sure if that's realistic or not, but thought I'd ask.

ptrikutam:
Good question. I don't really know, actually. [A quick google search](https://stackoverflow.com/questions/1339416/unit-testing-bash-scripts) reveals that there are things out there that do that. Perhaps something that we could do in a separate PR?

ptrikutam:
Looks like the solutions aren't great for unit testing shell scripts. Might add more complexity than is worth it to test what is currently a relatively simple script.

PR #217, opened 2016-02-08
Fix sidekiq in development

joshsmith:
This removes a mistake in the `Gemfile` whereby I placed `sidekiq-testing` in `development` group.


joshsmith:
This also adds the ability to view Sidekiq's dashboard in development, and adds a Sidekiq configuration file. The `paperclip` queue is not used, but it may be if we decided to use delayed paperclip.

PR #219, opened 2016-02-09
Add preview fields for body and markdown

begedin:
Not done yet.

As expected, there are tweaks for things we've missed, now that I'm working on the ember part of it.
- `title` should not be required for posts, if the post is a draft, otherwise, preview wont work
- the `active` scope was broken on posts, and the post index action was returning only `published`, not `edited` posts
- the `active` scope was completely missing on comments
- the markdown rendering + preview/publish flow was clashing, so I attempted to simplify it a bit. this should make it faster, simpler to ready and a bit less bug prone, I hope, but it does mean we had to remove the rendering method from the before_validation hook and we're now running it manually.

### DIFF
--- a/app/policies/post_policy.rb
+++ b/app/policies/post_policy.rb
@@ -15,40 +15,36 @@ class PostPolicy
   end
 
   def create?
-    # Cannot create if there's no user
-    return unless user.present?
+    # unauthenticated user cannot create
+    return false unless user.present?
 
-    # Cannot create if not the user who's posting
+    # user cannot create posts for other users
     return false unless post.user == user
 
-    # Can create issue and idea posts for any user
-    return true if post.issue? or post.idea?
+    # non-contributors can create issues or ideas
+    return true if post.issue?
+    return true if post.idea?
 
-    # Cannot create if not a contributor
-    return false unless organization_member_for_user
-
-    # Can create if a contributor with permissions
+    # contributors can create all types of posts
     return true if current_user_is_at_least_contributor_to_organization?
   end
 
   def update?
-    # Cannot update if there's no user
-    return unless user.present?
+    # unauthenticated user cannot update
+    return false unless user.present?
 
-    # Can update if the user is a project admin
+    # admin can update any post, even that belonging to other users
     return true if current_user_is_at_least_admin_in_organization?
 
-    # Cannot update if not the user who posted
-    return false unless post.user == user
-
-    # Can create issue posts for any user
-    return true if post.issue?
-
-    # Can create if a contributor with permissions
-    return true if current_user_is_at_least_contributor_to_organization?
+    # user can update their own post
+    return true if post.user == user
   end
 
   private
+
+    def user_is_member_of_organization
+      organization_member_for_user.present?
+    end
 
     def organization_member_for_user
       @organization_member_for_user ||= OrganizationMembership.find_by(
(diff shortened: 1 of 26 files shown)


joshsmith:
I fixed the failing spec by renaming `post` in a `let` to `users_post`. Not ideal, and very strange behavior.

When I did a `binding.pry` into the `authenticated_post` it appeared that `post` was an instance of `Post`. This is really strange to me, as I'm not sure how that data would pass through there.

If you can think of better naming or a better fix, please do. Mine is temporary and just meant to not leave you in a lurch. I would open an issue for it, if I were you.

begedin:
@JoshSmith I knew it'd be something weird. I guess it's something to do with how the `let` blocks are getting evaluated. Hound keeps telling me not to rely on instance variables and instead use `let` blocks, but this might be one of those cases where instance variables are slightly safer. 

Still, it might just be something with ordering.

begedin:
This is possibly done, outside of hound issues, but I'd like to make sure the ember part is merged first, before we're sure we haven't missed anything here.

begedin:
I need to merge develop into here, and clean it up. Once that's done, we can consider it mergeable together with the preview flow PR on the ember side.

joshsmith:
@begedin what kind of clean up needs to happen?

begedin:
@JoshSmith Just hound errors and then a quick scan of the changes to make sure I didn't leave anything bad in there.

joshsmith:
I can do that, then.

joshsmith:
I did a bunch of Hound fixes here except for all the line too long issues.

I still need to do a code review on this.

begedin:
@JoshSmith: I rebased/squashed into a single commit and looked through the code one last time. Feel free to review and merge if you consider it good.

spec/serializers/post_serializer_spec.rb
```diff
@@ -84,6 +84,14 @@
         expect(subject["markdown"]).to eql resource.markdown
       end
 
+      it "has a 'body_preview'" do
+        expect(subject["body_preview"]).to eql resource.body_preview
+      end
+
+      it "has 'markdown_preview'" do
+        expect(subject["markdown_preview"]).to eql resource.markdown_preview
+      end
```
joshsmith, line 97: I would not in `.to_not be_nil` expectations for all of these that can have them. We've been bitten by this before, where it's `nil` in cases and matches the model's `nil` view.

spec/serializers/comment_serializer_spec.rb
```diff
@@ -63,6 +63,14 @@
         expect(subject["markdown"]).to eql resource.markdown
       end
 
+      it "has a 'body_preview'" do
+        expect(subject["body_preview"]).to eql resource.body_preview
+      end
+
+      it "has 'markdown_preview'" do
+        expect(subject["markdown_preview"]).to eql resource.markdown_preview
+      end
```
joshsmith, line 75: Same thing here with `nil`.

app/serializers/post_serializer.rb
```diff
@@ -21,7 +21,8 @@
 #
 
 class PostSerializer < ActiveModel::Serializer
-  attributes :id, :title, :body, :status, :post_type, :likes_count, :markdown,
+  attributes :id, :title, :body, :body_preview, :status, :post_type,
+             :likes_count, :markdown, :markdown_preview,
```
joshsmith: Stylistic note, but I would group the relevant attributes together if we're not doing it alphabetically.

app/policies/post_policy.rb
```diff
@@ -15,41 +15,36 @@ def show?
...
+    # unauthenticated user cannot create
+    return false unless user.present?
 
-    # Cannot create if not the user who's posting
+    # user cannot create posts for other users
     return false unless post.user == user
 
-    # Can create issue and idea posts for any user
+    # non-contributors can create issues or ideas
     return true if post.issue? or post.idea?
```
joshsmith: We might want to split this onto two separate lines for clarity.

joshsmith:
This looks really good. There are some merge conflicts now. Made some minor comments, but nothing terribly blocking.

PR #249, opened 2016-02-24
Fixes for organization profile

joshsmith:
- [x] Allow for projects show
- [x] Test for `projects/:id` in API specs
- [x] Add projects to unavailable slugs if not added
- [x] Ensure tests don't include owner
- [x] Test for slugged route serializer
- [x] Test for `has_many` projects in organization
- [x] Test for `has_many` members in organization
- [x] Test for slug in project serializer

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ group :development, :test do
   gem 'pry-rails'
   gem 'pry'
   gem 'pry-nav'
+  gem 'pry-remote'
   gem 'pry-stack_explorer'
 
   gem 'rspec-rails', '~> 3.0'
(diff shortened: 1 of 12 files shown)


scarfdestroyer:
So @JoshSmith Now, "projects" is a reserved route that redirects to all projects and not, say, code-corps project called "projects" ?

scarfdestroyer:
Looks good to me. :+1:

joshsmith:
@scarfdestroyer actually `projects` was always in reserved routes. I just verified it was in there. The difference in this PR is just that `projects/:id` show works where it didn't before. This is required on the Ember side when doing an async request for a single project inside an organization's list of projects.

PR #240, opened 2016-02-18
Add organization description

scarfdestroyer:

PR #250, opened 2016-02-25
Fix posts and organizations

joshsmith:
This fixes issues with posts not loading all the included data (so usernames weren't showing up). https://github.com/code-corps/code-corps-ember/issues/100

This also fixes an error I missed in my code review of #240.


scarfdestroyer:
Looks good as far as I can see.

Issue #243, opened 2016-02-19
Make sure email is not exposed from API unless publicly available


joshsmith:
This issue includes two things:
- Allowing a user an option to publicly expose their email address
- Ensuring email is not provided by the API unless they either A) have the access rights, or B) it is publicly exposed

joshsmith:
I'm going to simplify this and say the email display to the public should not be a configurable option right now. Let's just keep it private by default and only expose to an authenticated user.

PR #251, opened 2016-02-25
Fix authenticated user serializer

joshsmith:
This adds `has_many` relationships missing from the authenticated user.


begedin:
Changes look good, but looking at this right now, is there much of a point to keep `user_serializer` and  `authenticated_user_serializer`, especially considering how Ember behaves.

The differences are that `user_serializer` has `facebook_id` and `facebook_access_token`, while `authenticated_user_serializer` does not.

joshsmith:
I think it's a good idea to keep it separate, because we don't want some of those things exposed in the general case. The gap may widen as time goes on, too.

begedin:
I understand and generally, I think this is good to go :+1: 

However, It feels sort of weird that the `user_serializer` actually provides more information than the `authenticated_user_serializer`.

joshsmith:
Ah, good catch! I think those are supposed to reversed, actually.

joshsmith:
No, these can't be reversed. I think I've changed my mind and we should just remove this extra serializer altogether.

joshsmith:
If we're going to hide certain attributes, like I want to do in #243, then we should be doing that all in the same serializer. These things might get pretty granular, so we may as well have a single point of failure for it.

begedin:
:+1:  Looks good to me. I'm thinking, if we need to diverge, we re-add the separation.

Overall, I think responses should mostly only differentiate in includes, not fields or relationships, unless there's a specific need for some sort of optimization

joshsmith:
Right, like there is a need for that optimization in `PostUserMentionsSerializer`, which does not need the entire `Post` serialized with all its includes.

PR #252, opened 2016-02-25
Add status field to user mentions. Change behavior so only old mentions of the same status are destroyed on regeneration

begedin:
This is done and the specs cover the behavior, but I would like to check for some refactoring potential before merging tomorrow. Here's what we have:
- `GET /comment_user_mentions` endpoint
- `GET /post_user_mentions` endpoint
- both endpoint accept a `comment/post_id` and a `status` parameter
- neither of the endpoints requires logging in
- modified behavior of scenarios for generating user mentions
  - if we're publishing a post, mentions are generated from `body` and old mentions with status `published` are being destroyed.
  - if we're previewing a post, mentions are generated from `body_preview` and old mentions with status `preview` are being destroyed
- the `update` method has been modified further. It now calls the generation of mentions explicitly, instead of this being left to `after_save`. The reason for that is, we have to pass in a boolean to indicate if we're doing it for a publish or a preview. **This is the main thing I would like to rewrite in a nicer way**.

### DIFF
--- a/app/controllers/comment_user_mentions_controller.rb
+++ b/app/controllers/comment_user_mentions_controller.rb
@@ -1,0 +1,16 @@
+class CommentUserMentionsController < ApplicationController
+  def index
+    authorize CommentUserMention
+    mentions = CommentUserMention.includes(:user, :comment).where(filter_params)
+    render json: mentions
+  end
+
+  private
+
+    def filter_params
+      {
+        comment_id: params[:comment_id],
+        status: params[:status]
+      }
+    end
+end
--- a/app/controllers/post_user_mentions_controller.rb
+++ b/app/controllers/post_user_mentions_controller.rb
@@ -1,0 +1,16 @@
+class PostUserMentionsController < ApplicationController
+  def index
+    authorize PostUserMention
+    mentions = PostUserMention.includes(:user, :post).where(filter_params)
+    render json: mentions
+  end
+
+  private
+
+    def filter_params
+      {
+        post_id: params[:post_id],
+        status: params[:status]
+      }
+    end
+end
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -37,7 +37,7 @@ class CommentsController < ApplicationController
 
     authorize comment
 
-    if comment.update publish?
+    if comment.update(publish?)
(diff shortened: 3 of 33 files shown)


joshsmith:
Moving this out of `after_save` seems like a dangerous thing to do, given that we used that lifecycle hook to explicitly ensure that generated user mentions were wrapped in a transaction. If the generation of these mentions failed, we would get a record rollback.

Is there no way to keep this method inside `after_save`?

begedin:
I'm thinking of options for that, but the difficulty is that us us generating a preview or a body isn't tied into any db column. Perhaps an instance variable for the model would work. I'll give it a go.

begedin:
I managed to refactor the update flow by using an instance variable indicating if the post is being saved for publishing, or for preview, to the point where it should be much safer.

Now the flow is

``` Ruby
before_validation :render_markdown_to_body
before_validation :publish_changes # skipped if not publishing
after_save :generate_mentions
```

`publish_changes` sets a new post/comment state and copies the contents of `_preview` fields into the regular fields.

app/controllers/comment_user_mentions_controller.rb
```diff
@@ -0,0 +1,16 @@
+class CommentUserMentionsController < ApplicationController
+  def index
+    authorize CommentUserMention
+    mentions = CommentUserMention.includes(:user, :comment).where(filter_params)
```
joshsmith, line 4: I'm fairly certain the `authorize` should be on the `mentions`, not the model itself.

app/controllers/post_user_mentions_controller.rb
```diff
@@ -0,0 +1,16 @@
+class PostUserMentionsController < ApplicationController
+  def index
+    authorize PostUserMention
+    mentions = PostUserMention.includes(:user, :post).where(filter_params)
```
joshsmith, line 4: Same point about the `authorize` here.

app/controllers/comment_user_mentions_controller.rb
joshsmith, line 4: I could be wrong about this so correct me if I'm wrong.

joshsmith, line 4: I am very wrong here.

PR #255, opened 2016-02-27
Make slugs case insensitive

### LINKED ISSUE #247: Make slugs case-insensitive
Right now there's a distinction between `joshsmith` and `JOSHSMITH`, which means someone could ostensibly claim both URLs.

We need to fix this case sensitivity in a way that ensures `JOSHSMITH` resolves to the same URL.

joshsmith:
Closes #247.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -111,7 +111,9 @@ class ProjectsController < ApplicationController
 
     def find_project_with_slugged_route!
       slugged_route = find_slugged_route!
-      Project.includes(:github_repositories, :organization).find_by!(slug: project_slug, organization: slugged_route.owner)
+      Project.includes(:github_repositories, :organization).
+        where("lower(slug) = ?", project_slug.downcase).
+        find_by!(organization: slugged_route.owner)
     end
 
     def find_projects_with_slugged_route!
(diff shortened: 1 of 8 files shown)


joshsmith:
This could maybe be refactored slightly into a query object, or at least scopes on both `Project` and `SluggedRoute`.

app/controllers/projects_controller.rb
```diff
@@ -111,7 +111,10 @@ def project_show?
 
     def find_project_with_slugged_route!
       slugged_route = find_slugged_route!
-      Project.includes(:github_repositories, :organization).find_by!(slug: project_slug, organization: slugged_route.owner)
+      Project.includes(:github_repositories, :organization).
+        where("lower(slug) = ?", project_slug.downcase).
+        where(organization: slugged_route.owner).
+        first!
```
begedin: Would 

``` ruby
where("lower(slug) = ?", project_slug.downcase).
find_by!(organization: slugged_route.owner)
```

not work here?

joshsmith: I'm not totally sure. I can try it.

joshsmith: It does. Pushing up change.

begedin:
:+1: Looks good to me.

PR #263, opened 2016-03-18
Allow users to join projects

begedin:
This is just a bunch of pending specs right now. I wasn't able to do anything else in time.

It would be good to get an opinion on this, though, since I'm not even sure this is the behavior we would like.

### DIFF
--- a/app/controllers/organization_memberships_controller.rb
+++ b/app/controllers/organization_memberships_controller.rb
@@ -1,0 +1,40 @@
+class OrganizationMembershipsController < ApplicationController
+  def index
+    render json: organization_memberships,
+           meta: meta_for(organization_membership_count),
+           each_serializer: OrganizationMembershipSerializer
+  end
+
+  def create
+  end
+
+  def update
+  end
+
+  private
+
+    def filter_params
+      filter_params = {}
+      filter_params[:organization] = organization
+      filter_params[:role] = params[:role] if params[:role]
+      filter_params
+    end
+
+    def organization_memberships
+      pp params
+      OrganizationMembership
+        .includes(:member)
+        .includes(:organization)
+        .where(filter_params)
+        .page(page_number)
+        .per(page_size)
+    end
+
+    def organization_membership_count
+      OrganizationMembership.where(filter_params).count
+    end
+
+    def organization
+      Organization.find(params[:organization_id])
+    end
+end
--- a/app/controllers/organization_memberships_controller.rb
+++ b/app/controllers/organization_memberships_controller.rb
@@ -52,12 +52,12 @@ class OrganizationMembershipsController < ApplicationController
     end
 
     def organization_memberships
-      OrganizationMembership
-        .includes(:member)
-        .includes(:organization)
-        .where(filter_params)
-        .page(page_number)
-        .per(page_size)
+      OrganizationMembership.
+        includes(:member).
+        includes(:organization).
+        where(filter_params).
+        page(page_number).
+        per(page_size)
     end
 
     def organization_membership_count
--- a/app/controllers/organization_memberships_controller.rb
+++ b/app/controllers/organization_memberships_controller.rb
@@ -1,17 +1,48 @@
 class OrganizationMembershipsController < ApplicationController
+  before_action :doorkeeper_authorize!, only: [:create, :update, :destroy]
+
   def index
+    authorize organization_memberships
     render json: organization_memberships,
            meta: meta_for(organization_membership_count),
            each_serializer: OrganizationMembershipSerializer
   end
 
   def create
+    organization_membership = OrganizationMembership.new(create_params)
+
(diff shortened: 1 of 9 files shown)


begedin:
This is now implemented. Missing:
- Policy specs - don't want to write them until we agree the current policy is fine
- Serializer specs
- Cleanup hound issues

I would focus on the policy file when reviewing.

joshsmith:
@begedin this is out of date with `development`. Is it ready for review? It looks like so.

begedin:
@JoshSmith This ought to be ready for review, but I'm not certain our policy is good here, so focus on that if you can.

app/controllers/organization_memberships_controller.rb
```diff
@@ -0,0 +1,92 @@
+class OrganizationMembershipsController < ApplicationController
+  before_action :doorkeeper_authorize!, only: [:create, :update, :destroy]
+
+  def index
+    authorize organization_memberships
+    render json: organization_memberships,
+           meta: meta_for(organization_membership_count),
+           each_serializer: OrganizationMembershipSerializer
+  end
```
joshsmith, line 10: Not entirely sure if this is index action is needed. What's the use case?

app/policies/organization_membership_policy.rb
```diff
@@ -0,0 +1,97 @@
...
+      user_is_member? && membership_for_user.owner?
+    end
+
+    def old_role_was_admin?
+      organization_membership.role_was == "admin"
+    end
+
+    def old_role_was_owner?
+      organization_membership.role_was == "owner"
+    end
```
joshsmith, line 90: I think we can just simplify to `role_was_admin?` and `role_was_owner?`

app/policies/organization_membership_policy.rb
```diff
@@ -0,0 +1,97 @@
...
+    true
+  end
+
+  def update?
+    # need to be logged in and member to do anything
+    return unless user_is_logged_in? && user_is_member?
+    # pending and contributors cant do any updates
+    return if user_is_contributor? || user_is_pending?
+    # admins cannot promote to admin or owner
+    return if user_is_admin? && (organization_membership.admin? || organization_membership.owner?)
```
joshsmith: I would think that an admin would be able to promote another admin, otherwise there's little utility to an admin here.

app/policies/organization_membership_policy.rb
```diff
@@ -0,0 +1,97 @@
...
+    # admins cannot demote from admin or owner
+    return if user_is_admin? && (old_role_was_admin? || old_role_was_owner?)
```
joshsmith: I think this means admins should be allowed to demote. Can you think of a counterexample?

app/policies/organization_membership_policy.rb
```diff
@@ -0,0 +1,97 @@
...
+    true
+  end
+
+  def destroy?
+    # need to be logged in and member to do anything
+    return unless user_is_logged_in? && user_is_member?
+    # owners and admins can destroy other roles, lesser users can destroy their own
+    return unless user_is_owner? || user_is_admin? || user_is_updating_own_role?
+    # admins cannot delete other admins
+    return if user_is_admin? && (old_role_was_admin? && !user_is_updating_own_role?)
```
joshsmith: I think admins should be able to remove other admins.

app/controllers/organization_memberships_controller.rb
joshsmith, line 10: Ah, this `index` is hit from `/organizations/:id/memberships`. This would be good to document here and not just leave it as an exercise for the reader to look in `routes.rb`. I think this is an important point.

joshsmith:
Just minor changes requested here. Let me know if you disagree with any of them. Otherwise, good work! 👍

app/policies/organization_membership_policy.rb
begedin: Generally speaking, I'm thinking users on the same credential levels should not be able to take each other's credentials. I think that should be reserved for users with higher credentials (owners, in this case).

I would say the same for promoting, but there, I can see the use if admins promoting others to admins.

I don't really disagree with your approach either, and this is easily changed, so I'm fine with merging.

It's a matter of philosophy, really. I'm imagining a scenario where there's a personal conflict between the admins. With your approach, the owner would be able to resolve issues that could happen. With my approach, there may be less of a chance for issues to happen, but then again, that could be at the cost of some flexibility.

I'll make the updates you requested and if there's ever a problem, we can revisit.

begedin:
@JoshSmith I realised I was completely missing the specs for the policy, so I had to write those. While on that, I noticed a few cases I didn't take into account, so I rewrote the policy a bit. Due to that, I'm thinking another review would be good. Feel free to rely on the specs now to get a better overview of the rules.

joshsmith:
👍

PR #272, opened 2016-05-19
Add Category

### LINKED ISSUE #267: Add Category model and associated models
An `InterestCategory` is something like "Politics" or "Technology." It's not just an interest because it's also a category for projects. This is the best naming convention I could think of it.

We also need `UserInterestCategory` and `ProjectInterestCategory`.models, both of which `belongs_to` each of those.

joshsmith:
Closes #267.
- [x] Add `Category` model and specs
- [x] Add categories API
- [x] Add `ProjectCategory` model and specs
- [x] Add `ProjectCategoryPolicy` and spec
- [x] Add `ProjectCategorySerializer` and probably one without includes, too
- [x] Add project categories API


joshsmith:
I moved some of this PR out into #274 instead because it was getting to be too large of a PR, and I wanted to do some refactoring in #273 first.

app/models/project_category.rb
```diff
@@ -0,0 +1,8 @@
+class ProjectCategory < ActiveRecord::Base
+  belongs_to :project
+  belongs_to :category
+
+  validates_presence_of :project
+  validates_presence_of :category
```
christopherstyles: These could be removed by adding `required: true` to the `belongs_to` associations above.

joshsmith: Didn't know that!

app/models/project_category.rb
```diff
@@ -0,0 +1,8 @@
...
+  validates_uniqueness_of :project_id, scope: :category_id
```
christopherstyles: [Style](https://github.com/bbatsov/rails-style-guide#sexy-validations):

> “sexy” validations.

app/models/category.rb
```diff
@@ -0,0 +1,17 @@
+class Category < ActiveRecord::Base
+  has_many :project_categories
+  has_many :projects, through: :project_categories
+
+  before_validation :add_slug_if_blank
+
+  validates :name, presence: true
+  validates :slug, presence: true
+  validates :slug, uniqueness: { case_sensitive: false }
+  validates :slug, slug: true
```
christopherstyles: Could these `slug` validations be [combined](https://github.com/bbatsov/rails-style-guide#sexy-validations)?

joshsmith: I'm not sure if they can with that custom validator, but I can try to do so.

app/models/project_category.rb
```diff
@@ -0,0 +1,6 @@
+class ProjectCategory < ActiveRecord::Base
+  belongs_to :project, required: true
+  belongs_to :category, required: true
+
+  validates :project_id, uniqueness: { scope: :category_id }
```
christopherstyles, line 5: May want to add a database constraint to back this up. Something along the lines of:

``` ruby
add_index(:project_categories, [:project_id, :category_id], unique: true)
```

christopherstyles, line 5: Also may want to add foreign key constraints that delete project categories if the project or category is removed…

``` ruby
add_foreign_key :project_categories, :projects, on_delete: :cascade
add_foreign_key :project_categories, :categories, on_delete: :cascade
```

joshsmith, line 5: @christopherstyles I don't imagine a `Project` or a `Category` ever being deleted, but still may not be a bad idea. Can we add this as a separate issue so we can do this for other models, too?

christopherstyles, line 5: Sure, it would probably be a good thing to get those missing constraints in the queue 👍

christopherstyles:
@begedin: Would you mind taking a glance at this one as well? I unfortunately ran out of time tonight, but most everything looks good so far. There were a few areas I could see being refactored/simplified (duplication in policies, finding by `id` or `slug` in controllers), but I think we can open separate issues for those, as they apply globally. 

Thanks ahead of time, very much appreciated!

app/controllers/project_categories_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    if project_category.project.blank?
+      project_category.valid?
+      render_validation_errors project_category.errors
+      return
+    end
+
+    authorize project_category
+
+    if project_category.valid?
```
begedin, line 16: I like this approach. Seems to be cleaner than `if object.save then` we usually do.

joshsmith, line 16: The `.blank?` check was only way I could reject things before it moved into Pundit. Not ideal.

spec/requests/api/categories_spec.rb
```diff
@@ -0,0 +1,94 @@
+require "rails_helper"
+
+describe "Categorys API" do
```
begedin: Categories

begedin:
Looks good, other than the tiny typo :+1:

PR #275, opened 2016-05-19
Update controllers with new methods for handling relationships

### LINKED ISSUE #273: Rewrite relationships in controllers to make use of deserialization
You can see usage of this in `ProjectCategoriesController` in #272, or just after it's merged.

joshsmith:
First stab at #273. Closes #273 if merged.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -68,12 +68,8 @@ class ApplicationController < ActionController::API
     ActiveModelSerializers::Deserialization.jsonapi_parse(params, options)
   end
 
-  def deserialized_params
-    parse_params(params)
-  end
-
   def require_param(key)
-    deserialized_params[key].presence || raise(ActionController::ParameterMissing.new(key))
+    parse_params(params)[key].presence || raise(ActionController::ParameterMissing.new(key))
   end
 
   def params_for_user(params)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -80,10 +80,6 @@ class ApplicationController < ActionController::API
     params.merge(user_id: current_user.id)
   end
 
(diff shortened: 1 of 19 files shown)


joshsmith:
I am open to any and all feedback here. I'm trying to make use of `ActiveModelSerializers::Deserialization.jsonapi_parse` which seems to cut down on a lot of the custom boilerplate we were writing in our own controllers.

I'm open to any feedback on whether this is a necessary change (although it seemed hard to do #272 without it), or if we need to improve on the syntax here. It's a pretty rough first pass.

app/controllers/application_controller.rb
```diff
@@ -64,6 +64,22 @@ def render_error(error)
     render json: error_hash, status: error_hash[:errors][0][:status]
   end
 
+  def parse_params(params, options = {})
+    ActiveModelSerializers::Deserialization.jsonapi_parse(params, options)
+  end
+
+  def deserialized_params
+    parse_params(params)
```
beauby: Here you could return `ActionController::Parameters.new(parse_params(params))`.

app/controllers/application_controller.rb
```diff
@@ -64,6 +64,22 @@ def render_error(error)
...
+  end
+
+  def require_param(key)
+    deserialized_params[key].presence || raise(ActionController::ParameterMissing.new(key))
```
beauby: And this would become `deserialized_params.require(key)`.

JoshSmith:
Testing.

begedin:
Looks good. I would just look into what @beauby is saying. Seems like it could be a bit cleaner.

joshsmith:
Yeah, I don't like the `require_param` thing at all. Hopefully his Rails PR goes through.

PR #277, opened 2016-05-20
Add ProjectRole

### LINKED ISSUE #270: Add ProjectRole model
We need to add a `ProjectRole` model and associated behavior.

joshsmith:
Closes #270.
- [x] Add `ProjectRole` model and specs
- [x] Add `ProjectRolePolicy` and specs
- [x] Add to `Project` as `has_many`
- [x] Add to `ProjectSerializer` and specs
- [x] Add `ProjectRoleSerialize` and specs
- [x] Add project role API specs, controller, routes, etc.

### DIFF
--- a/app/controllers/project_roles_controller.rb
+++ b/app/controllers/project_roles_controller.rb
@@ -1,0 +1,44 @@
+class ProjectRolesController < ApplicationController
+  before_action :doorkeeper_authorize!
+  before_action :require_params, only: [:create]
+
+  def create
+    project_role = ProjectRole.new(create_params)
+
+    if project_role.project.blank?
+      project_role.valid?
+      render_validation_errors project_role.errors
+      return
+    end
+
+    authorize project_role
+
+    if project_role.valid?
+      project_role.save!
+      render json: project_role, include: [:role, :project]
+    else
+      render_validation_errors project_role.errors
+    end
+  end
+
+  def destroy
+    project_role = ProjectRole.find(params[:id])
+
+    authorize project_role
+
+    project_role.destroy!
+
+    render json: :nothing, status: :no_content
+  end
+
+  private
+
+    def require_params
+      require_param :role_id
+      require_param :project_id
+    end
+
+    def create_params
+      parse_params(params)
+    end
+end
--- a/app/controllers/project_roles_controller.rb
+++ b/app/controllers/project_roles_controller.rb
@@ -5,15 +5,8 @@ class ProjectRolesController < ApplicationController
   def create
     project_role = ProjectRole.new(create_params)
 
-    if project_role.project.blank?
-      project_role.valid?
-      render_validation_errors project_role.errors
-      return
-    end
-
-    authorize project_role
-
     if project_role.valid?
+      authorize project_role
       project_role.save!
       render json: project_role, include: [:role, :project]
     else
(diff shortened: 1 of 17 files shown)


app/controllers/project_roles_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  before_action :require_params, only: [:create]
+
+  def create
+    project_role = ProjectRole.new(create_params)
+
+    if project_role.project.blank?
+      project_role.valid?
+      render_validation_errors project_role.errors
+      return
+    end
```
begedin: Is this because `project_id` can be an id that doesn't exist in the database, even though we require it as a parameter?

joshsmith: That's exactly right. So in one of the tests you can see that we're taking in an ID, but it's possible for it to go the policy and 500 because there's no organization for that project.

begedin: Hm. What would the validation message be in that case?

My guess is, there will be no errors at all, since the only validation is

``` Ruby
validates :project_id, uniqueness: { scope: :role_id }
```

While the spec passes (because it's an error response with the `VALIDATION_ERROR` id), the validation messages in the response would be blank.

Maybe we should add a 

``` Ruby
validates :project, presence: true
validates :role, presence: true
```

instead?

spec/requests/api/project_roles_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+      context "when there's no project with the specified id" do
+        it "fails with a validation error" do
+          authenticated_post "/project_roles", { data: { relationships: {
+            project: { data: { type: "projects", id: 55 } },
+            role: {
+              data: {
+                type: "roles", id: @role.id
+              }
+            }
+          } } }, token
```
joshsmith, line 101: @begedin this is where the error would happen in the policy. Nothing in those validations would help the policy determine this, because the validation hasn't been run before we hit the `authorize`.

app/models/project_role.rb
```diff
@@ -0,0 +1,6 @@
+class ProjectRole < ActiveRecord::Base
+  belongs_to :project, required: true
+  belongs_to :role, required: true
+
+  validates :project_id, uniqueness: { scope: :role_id }
```
begedin, line 5: Keep in mind that the error message generated here will be something along the lines of "Project_id must be unique, scoped to Role_id", which isn't exactly user friendly.

Maybe it should be

``` Ruby
validates :project, uniqueness: { scope: :role }
```

However, in that case, we may be doing db requests for records during validation. Also, the shoulda matchers have problems with that, I think. Maybe just providing a custom message using

``` Ruby
validates :project_id, uniqueness: { scope: :role_id , message: "Custom message" }
```

would do the trick?

joshsmith, line 5: I agree the error message is not the best, but this is really an edge case anyway. For the user who's adding a role to a project, it should probably just fail silently in the UI. I'm not sure we should spend a ton of time on it if it's not a really obvious problem. I'm guessing it's a 1/1000+ thing.

spec/requests/api/project_roles_spec.rb
begedin, line 101: I see your point. I still think we need the validation rule for presence, though, since otherwise, no validation message will be generated, just an empty validation error.

My vote is, we add the rules (role and project should both be present), then the controller flow is `validate > authorize > save`

joshsmith, line 101: @begedin the validations are present already. `belongs_to :project, required: true` does this.

begedin:
Looks good now 👍

PR #278, opened 2016-05-20
Add ProjectSkill

### LINKED ISSUE #271: Add ProjectSkill model
We need to add a `ProjectSkill` model and associated behavior.

joshsmith:
Closes #271.
- [x] Add `ProjectSkill` model and specs
- [x] Add `ProjectSkillPolicy` and specs
- [x] Add to `Project` as `has_many`
- [x] Add to `ProjectSerializer` and specs
- [x] Add `ProjectSkillSerializer` and specs
- [x] Add project skill API specs, controller, routes, etc.

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -1,3 +1,15 @@
+# == Schema Information
+#
+# Table name: categories
+#
+#  id          :integer          not null, primary key
+#  name        :string           not null
+#  slug        :string           not null
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#  description :text
+#
+
 class CategoriesController < ApplicationController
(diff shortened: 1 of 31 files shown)


begedin:
Looks good 👍 

I like that we aren't exposing "project_skills" with the project serializer, only "skills". We don't really need the weak entity client side and it simplifies things, probably. 

We should keep that in mind with other similar cases.

joshsmith:
@begedin actually I do think we might need the weak entity client-side when it comes to creating/deleting the project skill. We ended up needing it in the case of users categories and users skills.

begedin:
@JoshSmith Either way, it's something to keep in mind. It's possible that sometimes, we wont.

joshsmith:
Yes, agreed.

PR #280, opened 2016-05-21
Add project visibility and publishing

### LINKED ISSUE #265: A user can see a project in the list of published projects only if made public and provided enough details
A project cannot be seen users at large unless the admins have made it explicitly visible, and the project has at least a title, a description, a category, and a skill.

joshsmith:
Closes #265.

First stab. Really kind of gross in a lot of places, especially the testing story.

I'm also really not happy about the fact that there's no good way to show projects to organization admins only in an organization's page without explicitly checking for their admin status there. This is not yet implemented but I realized it was an edge case at the last moment.


joshsmith:
Closing as irrelevant because moving to Phoenix.

PR #281, opened 2016-05-21
Hide sensitive user data unless the user themselves is requesting it

### LINKED ISSUE #243: Make sure email is not exposed from API unless publicly available

joshsmith:
Fixes #243.

This is failing because the email no longer returns on `forgot_password`. This is also true on `create`.

As per discussion below, we need to separate out the reset password actions and also the fix the `create` action. Need to pick a path.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -71,7 +71,7 @@ class UsersController < ApplicationController
     user = User.find_by(email: forgot_password_params[:email])
 
     if user && user.forgot_password!
-      render json: user
+      render json: user, include_email: true
     else
       render_no_such_email_error
     end
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -45,4 +45,26 @@ class UserSerializer < ActiveModel::Serializer
   def photo_large_url
     object.photo.url(:large)
   end
+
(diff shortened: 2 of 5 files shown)


begedin:
For `create`, we might want to log in the user as soon as they are created, if the action is not done by an admin? That would be sort of common behavior, no?

For `forgot_password`, it depends. If we want to show something like `reset password details have been sent to #{email}` then we clearly need it, but if not, we could just not need it in the response. The message could just as easily be "reset password details have been sent to the email you provided". If we do need it, then maybe `forgot_password` should just use a different serializer. Realistically, our client should not be expecting a user model response during such an action anyway. It will probably be a manual ajax request of some sort, with a manual response handler.

joshsmith:
@begedin we could even use something like https://github.com/mike-north/ember-api-actions

joshsmith:
I do think the email as a response would be good here. It's better UI.

joshsmith:
I also think it would be preferable to log them in. This happens client-side manually somehow now. I don't think I get the flow.

joshsmith:
👍 thanks!

PR #283, opened 2016-05-22
Add organization membership show and more organization membership relationships needed on Ember side

joshsmith:
Pretty self-explanatory. Should all be fully tested. Just needs review.


begedin:
Looks good 👍

PR #285, opened 2016-05-25
Change post filtering

joshsmith:
Not even sure this is how we should be doing this, but it's a first go. No tests. One-line change.


begedin:
Seems fine to me. Not sure if you wanna tackle the hound warning or not. I'm fine either way.

PR #286, opened 2016-06-02
Add procfile

joshsmith:


Procfile
```diff
@@ -0,0 +1,2 @@
+web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-development}
```
christopherstyles, line 1: I’d probably recommend allowing the the `PORT` AND `RACK_ENV` to be configured via the environment, or even better, consider adding a config file for finer-grained control over all these options. The Heroku [example](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#config) is similar to what I usually set up.

PR #289, opened 2016-06-02
Add onboarding for users

joshsmith:
This adds a `state` and `state_transition` to `User` to be able to track onboarding progress in the front-end of the app.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,6 +27,7 @@
 class User < ActiveRecord::Base
   ASSET_HOST_FOR_DEFAULT_PHOTO = "https://d3pgew4wbk2vb1.cloudfront.net/icons".freeze
 
+  include AASM
   include Clearance::User
 
   has_many :organization_memberships, foreign_key: "member_id"
@@ -69,8 +70,36 @@ class User < ActiveRecord::Base
   validates :website, format: { with: /\A((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,}(([0-9]{1,5})?\/.*)?#=\z/ix }, allow_blank: true
 
   validate :slug_is_not_duplicate
+  validate :can_transition
 
+  before_save :attempt_transition
   after_save :create_or_update_slugged_route
+
+  attr_accessor :state_transition
+
+  # User onboarding
+  aasm do
+    state :signed_up, initial: true
+    state :selected_categories
+    state :selected_roles
+    state :selected_skills
+
+    event :select_categories do
+      transitions from: :signed_up, to: :selected_categories
+    end
+
+    event :select_roles do
+      transitions from: :selected_categories, to: :selected_roles
+    end
+
+    event :select_skills do
+      transitions from: :selected_roles, to: :selected_skills
+    end
+  end
+
+  def state
+    aasm_state
+  end
 
   # Follows a user.
   def follow(other_user)
@@ -88,6 +117,15 @@ class User < ActiveRecord::Base
   end
 
   private
+
+    def can_transition
+      return if state_transition.blank?
+      errors.add(:state, "cannot transition") unless send("may_#{state_transition}?")
+    end
+
+    def attempt_transition
+      send(state_transition) if state_transition.present?
+    end
 
     def slug_is_not_duplicate
       if SluggedRoute.where.not(owner: self).where("lower(slug) = ?", username.try(:downcase)).present?
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -75,6 +75,7 @@ class User < ActiveRecord::Base
   before_save :attempt_transition
   after_save :create_or_update_slugged_route
 
+  alias_attribute :state, :aasm_state
   attr_accessor :state_transition
 
   # User onboarding
@@ -95,10 +96,6 @@ class User < ActiveRecord::Base
     event :select_skills do
       transitions from: :selected_roles, to: :selected_skills
     end
-  end
-
-  def state
-    aasm_state
   end
 
   # Follows a user.
--- a/db/migrate/20160602211946_add_state_to_users.rb
+++ b/db/migrate/20160602211946_add_state_to_users.rb
@@ -1,5 +1,5 @@
 class AddStateToUsers < ActiveRecord::Migration
   def change
-    add_column :users, :aasm_state, :string, default: "signed_up"
+    add_column :users, :aasm_state, :string, default: "signed_up", null: false
   end
 end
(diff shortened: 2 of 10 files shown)


app/models/user.rb
```diff
@@ -69,9 +70,37 @@ class User < ActiveRecord::Base
...
+      transitions from: :selected_categories, to: :selected_roles
+    end
+
+    event :select_skills do
+      transitions from: :selected_roles, to: :selected_skills
+    end
+  end
+
+  def state
+    aasm_state
```
schneidmaster: Consider `alias state aasm_state` for simplicity.

db/migrate/20160602211946_add_state_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddStateToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :aasm_state, :string, default: "signed_up"
```
christopherstyles: May want to add a not null constraint on this, as it should probably never be empty.

app/models/user.rb
joshsmith: For some reason this breaks things in unexpected ways. Causes slug validation to behave wildly.

db/migrate/20160602211946_add_state_to_users.rb
joshsmith: Fixed.

app/models/user.rb
christopherstyles: Perhaps `alias_attribute :state, :aasm_state`?

spec/models/user_spec.rb
```diff
@@ -151,6 +152,29 @@
     end
   end
 
+  describe "onboarding" do
+    let(:user) { create(:user) }
+
+    it "transitions correctly when state_transition is set and saved" do
+      expect(user.state).to eq "signed_up"
```
christopherstyles: `aasm` also provides some nice [RSpec matchers](https://github.com/aasm/aasm#testing). The above could also be written:

``` ruby
 expect(user).to have_state(:signed_up)
```

I like the expressiveness, but either style works just the same.

spec/models/user_spec.rb
```diff
@@ -151,6 +152,29 @@
...
+
+      user.state_transition = "select_categories"
+      user.save
```
christopherstyles, line 162: An alternative to calling `save` after setting each transition by direct assignment would be to use auto-save with a bang:

``` ruby
user.select_categories!
# => true

user.aasm.current_state
# => :selected_categories
```

christopherstyles, line 162: Nevermind that, I see now you’re going through the `state_transition` accessor and not directly assigning. I need to go to sleep 😴

app/models/user.rb
```diff
@@ -69,9 +70,34 @@ class User < ActiveRecord::Base
...
 
+  alias_attribute :state, :aasm_state
+  attr_accessor :state_transition
+
+  # User onboarding
+  aasm do
+    state :signed_up, initial: true
+    state :selected_categories
+    state :selected_roles
+    state :selected_skills
```
christopherstyles, line 86: I wonder if these state names could benefit from becoming progressive tense, so that it’s unambiguous the user is currently in the state of taking an action. E.g.

``` ruby
state :signing_up, initial: true
state :selecting_categories
state :selecting_roles
state :selecting_skills
```

In turn, the events would change slightly:

``` ruby
# event :select_categories do
#   transitions from: :signed_up, to: :selected_categories
# end
#
# Becomes:
event :select_categories do
  transitions from: :signing_up, to: :selecting_categories
end
```

Having the state represent the current action a user is taking seems like it would be easier to reason about on the frontend. Having a state in the past tense introduces some cognitive overhead about what steps must have come before and what must come next, as opposed to what the user should be doing only _now_…an order dependency that may not be necessary. I guess it just depends on how you’re using the state names on the frontend, I’m not very familiar with the requirements there.

christopherstyles:
@JoshSmith Everything is looking good here, and functions as expected. 👍

app/models/user.rb
begedin, line 86: I agree with the idea, but in that case, the states would probably be

``` Ruby
state :selecting_categories, initial: true
state :selecting_roles
state :selecting_skills
state :all_done # or whatever the more applicable name would be
```

My reasoning is, there is no way for a user to be in the `signing_up` state. If the user record exist, then they already signed up.

begedin:
👍 looks good to me as well, but I think the states are worth discussing a bit.

app/models/user.rb
joshsmith, line 86: The reason I went with past tense was what @begedin said.

joshsmith, line 86: Here is the front-end, for context:

``` js
  currentStep: 1,
  totalSteps: 3,

  steps: [
    {
      number: 1,
      state: 'signed_up',
      currentRoute: 'start.interests',
      nextRoute: 'start.expertise',
    },
    {
      number: 2,
      state: 'selected_categories',
      currentRoute: 'start.expertise',
      nextRoute: 'start.skills',
    },
    {
      number: 3,
      state: 'selected_roles',
      currentRoute: 'start.skills',
      nextRoute: 'projects-list',
    }
  ],

  currentStep: Ember.computed('currentUser.user.state', function() {
    let state = this.get('currentUser.user.state');
    let steps = this.get('steps');

    return steps.find((step) => { return step.state === state});
  }),

  currentStepNumber: Ember.computed.alias('currentStep.number'),
  currentStepState: Ember.computed.alias('currentStep.state'),

  isSelectingCategories: Ember.computed.equal('currentStepState', 'signed_up'),
  isSelectingRoles: Ember.computed.equal('currentStepState', 'selected_categories'),
  isSelectingSkills: Ember.computed.equal('currentStepState', 'selected_roles'),

  isOnboarding: Ember.computed.or('isSelectingCategories', 'isSelectingRoles', 'isSelectingSkills'),

  progressPercentage: Ember.computed('currentStepNumber', 'totalSteps', function() {
    return (this.get('currentStepNumber') / this.get('totalSteps')) * 100;
  }),

  currentRoute: Ember.computed.alias('currentStep.currentRoute'),
  nextRoute: Ember.computed.alias('currentStep.nextRoute'),
```

joshsmith, line 86: One other thing of concern is if the first step changed. The initial state of `signed_up` was nice because that would _always_ be true, whereas we would have to consider some migrations of data in the event that first step did change.

begedin:
@JoshSmith After the context, I agree with your approach. I think this is good to merge.

PR #292, opened 2016-06-08
Strip certain attributes in user model

### LINKED ISSUE #51: Handle "" correctly in /settings/profile
Currently in /settings/profile if a user removes their name/website etc. from the the text box and then clicks update it will erase their previous data.

Not sure what we should be checking for/doing here. For example, we probably want them to be able to remove a website they no longer have access to or user but probably we probably want them to have a name once they add one. 
This or at least ask, "Are you sure you want to remove your Name?" if they previously had information and then are setting it to an empty string.

joshsmith:
This fixes https://github.com/code-corps/code-corps-ember/issues/51.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,8 @@ gem "pusher"
 
 gem "seed-fu"
 
+gem "strip_attributes"
+
 group :development, :test do
   gem "annotate"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -301,6 +301,8 @@ GEM
       actionpack (>= 4.0)
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
+    strip_attributes (1.8.0)
+      activemodel (>= 3.0, < 6.0)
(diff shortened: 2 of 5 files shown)


begedin:
This looks good and I agree with the idea, but I believe code-corps/code-corps-ember#51 is concerned with preventing the user from blanking out some fields in their profile, once they have a value (such as "name"). 

I would merge this, but we need to consider if the described issue is actually an issue.

joshsmith:
@begedin I originally asked for that issue to be opened and it was misunderstood. The issue, now that I understand it, is that sending `""` as a value should be handled as `nil` and not `""`. This was rearing its head in the Twitter handle case, where I'd get an `@` for a username and had nothing to actually show.

If someone unsets their name, that is their prerogative. Something like an email/username should not be unset, but there's no way to do that currently anyway.

begedin:
@JoshSmith In that case, this looks good to merge

PR #295, opened 2016-06-12
Add user availability routes

joshsmith:
This adds routes for checking if the username or email is available.


begedin:
Looks good. I would just update the description, so we don't later think we merged without specs.

PR #296, opened 2016-06-13
Add searchkick

joshsmith:
This is good to review.

It adds searchkick and returns the top five most related queries for autocompleting search results. Tests are designed to not test searchkick itself but that the searchkick API was called.


rjsamson:
👍 👍 👍

PR #298, opened 2016-06-14
Fix users availability endpoints to work case-insensitive

### LINKED ISSUE #297: Username and email availability are case sensitive
Discovered when testing signup with "joshsmith" and my actual username is "JoshSmith". Gave me an invalid error using the slug validator rather than an availability error.

joshsmith:
This fixes #297.


idelahoz:
Looks good 👍

PR #306, opened 2016-06-15
Add segment

### LINKED ISSUE #305: Add analytics
We need to add `segment-ruby` and track a number of behaviors.

joshsmith:
Closes #305 when finished.

Now:
- [x] Added profile picture from gravatar
- [x] Added user category
- [x] Added user role
- [x] Added user skill
- [x] Approved organization membership
- [x] Created comment
- [x] Created comment image
- [x] Created post (type)
- [x] Created post image
- [x] Edited comment
- [x] Edited post
- [x] Previewed existing comment
- [x] Previewed new comment
- [x] Published comment
- [x] Published post
- [x] Rejected organization membership
- [x] Removed organization membership
- [x] Removed user category
- [x] Removed user role
- [x] Removed user skill
- [x] Requested organization membership
- [x] Selected categories
- [x] Selected roles
- [x] Selected skills
- [x] Signed in
- [x] Signed up
- [x] Updated comment
- [x] Updated profile
- [x] Updated profile picture

Later:
- [ ] Created comment user mention
- [ ] Created post user mention
- [ ] Created project
- [ ] Created project category
- [ ] Created project role
- [ ] Created project skill
- [ ] Removed project category
- [ ] Removed project role
- [ ] Removed project skill
- [ ] Updated project

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ ruby "2.2.3"
 gem "rails", "4.2.5.1"
 
 gem "rails-api"
+gem "active_model_serializers", github: "rails-api/active_model_serializers"
 
 gem "puma"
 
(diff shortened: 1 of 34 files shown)


begedin:
@JoshSmith After a good amount of research, for now, I just decided to move the analytics code into a service object, only exposing an instance within the `ApplicationController#analytics` method. This should make the whole thing much easier to stub and test. Also, since we really need to identify with every action we track, I just made it so that the tracking methods call identify instead of us calling it manually.

We could go a step further here, if we decide that there will only ever be a single tracking call per controller action. In that case, I could write an `ActionSupport::Concern` module, which we can just include in the `ApplicationController`. We could then do something like

``` Ruby
track :create, with: :track_user_created
track :update, with: track_user_updated

# this one doesn't need special parameter handling, so we just delegate
delegate :track_user_created, to: :analytics

# this one needs special parameterhandling, so we need a method
def track_user_updated
  analytics.track_user_updated(update_params)
end
```

at the top of our controller.

joshsmith:
@begedin this makes sense. I don't think we'll only have one track per controller action since some of our actions are rather complex. I was assuming this was the route we'd have to take. It's a bit of a bummer that we can't stub directly, but oh well.

begedin:
Looks good 👍 

Unfortunate we need so many methods and calls, but I don't really see a way around it.

PR #310, opened 2016-06-16
Allow post type to be saved during draft preview

begedin:
Fixes code-corps/code-corps-ember#129

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -54,9 +54,10 @@ class PostsController < ApplicationController
 
     authorize post
 
-    post.assign_attributes(update_params)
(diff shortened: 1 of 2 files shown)


app/controllers/posts_controller.rb
```diff
@@ -70,10 +71,14 @@ def publish?
...
     end
 
-    def update_params
+    def update_params_for_published
       parse_params(params, only: [:markdown_preview, :title])
     end
 
+    def update_params_for_draft
+      parse_params(params, only: [:markdown_preview, :title, :post_type])
+    end
```
begedin, line 80: We could technically do "Inheritance" here, but we don't gain much from it.

joshsmith:
👍 LGTM.

Issue #314, opened 2016-06-17
Discussion: Simplify preview behavior on various records


begedin:
Test.

Right now, we have a pretty complicated set of behaviors for rendering markdown for previews, where we have separate _markdown, _markdown_preview, _html and _html_preview fields and then there's logic that determines what goes into which field depending on the API request parameters.

My suggestion is the following.

We eliminate this behavior completely. We just keep `_markdown` and `_html` fields. No extra `_preview_` fields.

When the client decides to save the record, be it post, comment or project (with soon to be added `long_description`), the markdown field is taken and the html field is generated.

For the preview action, regardless of record, we have a single `generate_html` endpoint, which the client can send the markdown to and receive the HTML as well as a set of mentions.

Basically, the client behavior would be
- when the user requests a preview, we make a manual ajax request to the preview endpoint
- when the user saves the record, we set the record.markdown field and request a regular save

This would allow us to
- eliminate the custom adapter logic for posts, comments (and soon projects)
- simplify the save logic a tiny bit
- improve on the encapsulation of the `editor-with-preview` component

This would simplify the logic greatly and also allow us consolidate html and mention rendering.

The major reason why I think this would be an improvement is that I contacted github to see how they are handling it and the reply was surprising:

> I don't think our drafts persist. We used to have a feature that was keeping drafts around forever, but we removed that feature. We may bring it back if we can make it work properly.
> 
> As for the way that could work, there are probably quite a few different approaches. I wouldn't be surprised if there are components out there that do that for you. Without the requirement to reinvent the wheel.

Basically, their drafts and previews do not really persist. What is likely happening is that the browser somehow keeps the textarea filled in.

I think this is worth the discussion at the very least.

joshsmith:
I think they must be misinformed about this. Type a comment in here. Hit preview. Hit write. Refresh the page. I'm not sure how this would just be local storage. Maybe it is. I'm not sure.

christopherstyles:
In the past I’ve used [marked](https://github.com/chjj/marked) to render previews on the client, saving a trip to the server. This would be an additional frontend dependency though, which I’m assuming you’re trying to avoid.

Is it necessary to store the rendered `_html` alongside the markdown `body` in the db? Would it be possible to keep only the `body` (in markdown), and render it to html as-needed? A computed `_html` attribute could be added to the API response, which should be cached eventually so the performance penalty for rendering it each time would be negated.

[...]

PR #315, opened 2016-06-17
Allow post type to be edited

### LINKED ISSUE #312: Allow post types to be changed

begedin:
Closes #312 

Requires client changes as well, but can be merged without it.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -54,8 +54,7 @@ class PostsController < ApplicationController
 
     authorize post
 
-    post.assign_attributes(update_params_for_draft) if post.draft?
-    post.assign_attributes(update_params_for_published) unless post.draft?
+    post.assign_attributes(update_params)
 
     if post.update(publish?)
       GeneratePostUserNotificationsWorker.perform_async(post.id) if publish?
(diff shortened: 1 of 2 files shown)


app/controllers/posts_controller.rb
```diff
@@ -71,12 +70,9 @@ def publish?
...
 
-    def update_params_for_published
-      parse_params(params, only: [:markdown_preview, :title])
-    end
-
-    def update_params_for_draft
-      parse_params(params, only: [:markdown_preview, :title, :post_type])
+    def update_params
+      return parse_params(params, only: [:markdown_preview, :title]) unless publish?
+      return parse_params(params, only: [:markdown_preview, :title, :post_type]) if publish?
```
joshsmith, line 75: I think this changed behavior we had just fixed. When previewing the `post_type` gets unset. I don't think we need the `publish?` check here. `update_params` is the same for both states. Tests will need changed, too.

begedin, line 75: You still don't want to change the post type while requesting a preview. Otherwise, the user might give up on editing and then the post type remains changed. This is one of the several reasons why I wrote #314

joshsmith, line 75: Doesn't the post type get reset though? That's my concern. Steps to reproduce:
- You're on the new post form. You change from "Issue" to "Idea".
- You hit preview.
- You hit edit.
- The post type is changed back without you expecting it.

begedin, line 75: It is, which is why, with things as they are now, we need to modify that behavior as well, to make it work like the post title does, where we display a clone property and then only set the actual `post.post_type` when saving the post.

joshsmith:
I believe this is probably blocked right now by our other changes we need to make, but would it be good to close this now anyway @begedin? It doesn't work how we want it to, but also doesn't make sense maybe to hold this PR hostage. We'll fix it elsewhere in one go.

joshsmith:
This may need to be closed outright if #326 is merged.

joshsmith:
Closed by #326.

PR #316, opened 2016-06-17
Remove progress post types

### LINKED ISSUE #311: Remove progress post types

begedin:
Closes #311 

Anything else needed here?

How will we replace them with "updates" we've mentioned? Will it be a whole new type of record?

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -65,7 +65,6 @@ class Post < ActiveRecord::Base
 
   enum post_type: {
     idea: "idea",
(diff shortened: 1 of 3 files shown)


joshsmith:
I think it will be a `ProjectUpdate` record. Would love to get it into 1.0 but I think it needs more thought as of right now. Since it's just Code Corps on here for now, we can simply use the mailing list as a means of doing this.

joshsmith:
👍

PR #322, opened 2016-06-20
Allow post filtering by status

### LINKED ISSUE #321: Allow post filtering by status
Required for code-corps/code-corps-ember#214

The controller doesn't currently support this filter.

begedin:
Closes #321 

I wrote it so the API doesn't make any assumptions about default status. Assuming an "open" status would make it impossible to fetch both closed and open posts.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -83,6 +83,7 @@ class PostsController < ApplicationController
     def filter_params
       filter_params = {}
       filter_params[:post_type] = params[:post_type].split(",") if params[:post_type]
+      filter_params[:status] = params[:status] if params[:status]
       filter_params
     end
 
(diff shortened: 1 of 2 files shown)


joshsmith:
@begedin Can you explain this a bit more?

> Assuming an "open" status would make it impossible to fetch both closed and open posts.

begedin:
If our controller only returns "open" issues by default (when no filter is specified by the client), then we can't fetch "all" issues from the API, unless we have special behavior of some sort implemented for that specifically. 

Instead, our controller makes no assumptions. If status is not specified, it returns all issues. If it is specified, it returns issues of that status.

The client is the one that decided to display open issues by default (by fetching open issues from the API) and can fetch all issues by removing this default filter. This is generally how github behaves. By default, the client filters by `is:open`, but the user can remove that filter if they want to.

begedin:
@JoshSmith Should be good to review again.

christopherstyles:
👍 :shipit:

PR #324, opened 2016-06-21
Fix setup process

schneidmaster:
- Remove post fixture with deprecated "progress" type
- Remove schema.rb attribute that no longer exists in models


joshsmith:
👍

PR #326, opened 2016-06-21
Simplify preview behavior

### LINKED ISSUE #314: Discussion: Simplify preview behavior on various records

joshsmith:
This closes #314.

This adds a `/previews` endpoint where you can create a `Preview` model which `has_many :preview_user_mentions`. It takes in markdown and renders out HTML.

We're getting rid of the `_preview` fields on `Post` and `Comment`, as well as the concept of draft states.

The assumption here is that the client requests a `Preview` for any kind of record: it doesn't matter what that record is. Instead, we just render mentions and the filter the markdown through a pipeline.

Thus, previews are completely isolated from the behavior of objects we might want previews for.

This will allow the client to request any number of markdown-ified previews it wants. We could modify the preview to not generate mentions if we want, or do other custom things dependent on the preview. We could also add a polymorphic relationship and track what we generated a preview for.

This can't really be merged without a sister PR on the Ember side to make this work.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -16,8 +16,6 @@
 #  number           :integer
 #  aasm_state       :string
 #  comments_count   :integer          default(0)
-#  body_preview     :text
-#  markdown_preview :text
 #
 
 class PostsController < ApplicationController
@@ -41,8 +39,9 @@ class PostsController < ApplicationController
 
     authorize post
 
-    if post.update(publish?)
-      GeneratePostUserNotificationsWorker.perform_async(post.id) if publish?
+    if post.save
+      post.reload
+      GeneratePostUserNotificationsWorker.perform_async(post.id)
       render json: post
     else
       render_validation_errors post.errors
@@ -54,11 +53,11 @@ class PostsController < ApplicationController
 
     authorize post
 
-    post.assign_attributes(update_params_for_draft) if post.draft?
-    post.assign_attributes(update_params_for_published) unless post.draft?
+    post.assign_attributes(update_params)
 
-    if post.update(publish?)
-      GeneratePostUserNotificationsWorker.perform_async(post.id) if publish?
+    if post.save
+      post.reload
+      GeneratePostUserNotificationsWorker.perform_async(post.id)
       render json: post
     else
       render_validation_errors post.errors
@@ -71,17 +70,13 @@ class PostsController < ApplicationController
       true unless parse_params(params).fetch(:preview, false)
     end
 
-    def update_params_for_published
-      parse_params(params, only: [:markdown_preview, :title])
-    end
-
-    def update_params_for_draft
-      parse_params(params, only: [:markdown_preview, :title, :post_type])
+    def update_params
+      parse_params(params, only: [:markdown, :title, :post_type, :state])
     end
 
     def create_params
       params_for_user(
-        parse_params(params, only: [:markdown_preview, :title, :post_type, :project])
+        parse_params(params, only: [:markdown, :title, :post_type, :project])
       )
     end
 
@@ -113,7 +108,7 @@ class PostsController < ApplicationController
     def find_posts!
       project = find_project!
 
-      Post.active.
+      Post.
         includes(:user).
         includes(:project).
         includes(comments: :user).
@@ -125,6 +120,6 @@ class PostsController < ApplicationController
     end
 
     def post_count
-      Post.active.where(filter_params.merge(project_id: project_id)).count
+      Post.where(filter_params.merge(project_id: project_id)).count
     end
 end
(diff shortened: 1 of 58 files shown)


db/migrate/20160621031228_remove_status_from_user_mentions.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveStatusFromUserMentions < ActiveRecord::Migration
+  def change
+    remove_column :comment_user_mentions, :status
+    remove_column :post_user_mentions, :status
+  end
+end
```
christopherstyles, line 6: May want to add types and options here so the migration is reversible:

``` ruby
def change
  remove_column :comment_user_mentions, :status, :string, null: false, default: "preview"
  remove_column :post_user_mentions, :status, :string, null: false, default: "preview"
end
```

db/migrate/20160621003916_create_preview_user_mentions.rb
```diff
@@ -0,0 +1,14 @@
+class CreatePreviewUserMentions < ActiveRecord::Migration
+  def change
+    create_table :preview_user_mentions do |t|
+      t.belongs_to :user, null: false
```
christopherstyles: With an index + foreign key: `t.belongs_to :user, index: true, foreign_key: true, null: false`

db/migrate/20160621003916_create_preview_user_mentions.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.belongs_to :preview, null: false
```
christopherstyles: With an index + foreign key: `t.belongs_to :preview, index: true, foreign_key: true, null: false`

db/migrate/20160621003651_create_previews.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePreviews < ActiveRecord::Migration
+  def change
+    create_table :previews do |t|
+      t.text :body, null: false
+      t.text :markdown, null: false
+
+      t.belongs_to :user
```
christopherstyles: With an index + foreign key: `t.belongs_to :user, index: true, foreign_key: true`

config/routes.rb
```diff
@@ -24,13 +24,16 @@
 
     get "ping", to: "ping#index"
 
+    post "posts/generate_html", to: "posts#generate_html"
     resources :posts, only: [:create, :update] do
       resources :comments, only: [:index]
     end
```
christopherstyles, line 29: An alternate way of doing this, using the DSL, could be:

``` ruby
resources :posts, only: [:create, :update] do
  post :generate_html, on: :collection
end
```

or with a block:

``` ruby
resources :posts, only: [:create, :update] do
  collection do
    post :generate_html
  end
end
```

I find it a little easier to scan when nested this way, but it’s all [mostly] functionally equivalent.

app/controllers/posts_controller.rb
```diff
@@ -54,34 +53,34 @@ def update
...
+      post.reload
+      GeneratePostUserNotificationsWorker.perform_async(post.id)
       render json: post
     else
       render_validation_errors post.errors
     end
   end
 
+  def generate_html
+    Post.generate_preview(params[:markdown])
```
christopherstyles: Probably missing an obvious thing, but I don’t see this class method defined on `Post`?

christopherstyles:
This is a fantastic refactor, definitely a step in the right direction. Great work @JoshSmith!! 👏

app/controllers/posts_controller.rb
joshsmith: This is a good catch. Left over from the spike.

PR #327, opened 2016-06-21
Add newrelic_rpm to gems and configure

### LINKED ISSUE #300: Add New Relic

christopherstyles:
The New Relic [addon](https://elements.heroku.com/addons/newrelic) has been provisioned for the Heroku app, and a `NEW_RELIC_LICENSE_KEY` environment variable added to the configuration. 

New configuration for the app resides in `config/newrelic.yml`.

Resolves #300

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem "html-pipeline"
 gem "html-pipeline-rouge_filter"
 gem "kaminari"
 gem "koala"
+gem "newrelic_rpm"
 gem "obscenity"
 # paperclip master currently doesn"t work with new version of AWS SDK
 gem "paperclip", git: "https://github.com/thoughtbot/paperclip", ref: "523bd46c768226893f23889079a7aa9c73b57d68"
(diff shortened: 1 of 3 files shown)


joshsmith:
@christopherstyles has it been provisioned for `code-corps-staging` only or also the production app?

joshsmith:
👍 🐑 🇮🇹

christopherstyles:
@JoshSmith Oh that’s right – mind adding me to the staging app so I can set it up there as well?

joshsmith:
@christopherstyles sure, did you want to add something else to this config then?

christopherstyles:
@JoshSmith No need to add anything else, so long as the `RAILS_ENV` on the staging app is set to `production`. It’ll use the same environment var.

joshsmith:
@christopherstyles it is, just wasn't sure if it would be confusing which New Relic dashboard we're looking at (with naming). But probably doesn't matter all that much.

christopherstyles:
@JoshSmith Okay, gtg on staging now. Since there is no `staging` environment, it can’t be easily changed from the `newrelic.yml`. However, you should be able to change the name of the app from the “Account Settings” in the New Relic dashboard (you’ll need to, as only the app owner can do that).

joshsmith:
@christopherstyles awesome, thanks! 👍

PR #332, opened 2016-06-21
Add long description to projects

### LINKED ISSUE #313: Add a long_description to Project
We need to add a `text` `long_description` to `Project` that takes and renders markdown. This is where project owners will give detailed information about their project like:
- What do you hope to accomplish?
- What real problem is this solving?
- What kind of support will you need, both right now and in the future?
- What risks and challenges will you face?

This can be a field on `Project` itself. We likely need to add `long_description_markdown`.

Unlike a `Post` or `Comment`, this will not generate any user mentions.

joshsmith:
Closes #313.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -77,11 +77,17 @@ class ProjectsController < ApplicationController
   private
 
     def create_params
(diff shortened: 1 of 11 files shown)


begedin:
:+1:

PR #337, opened 2016-06-22
Fix Project#long_description_body not updating when Project#long_desription_markdown is saved as empty string

### LINKED ISSUE #336: When Project#long_description_markdown is blank, the field is saved, but _body is not updated
`Project` does not have a `presence` validation for `long_description_body` or `long_description_markdown`, which makes sense, since descriptions aren't present initially.

However, due to a bad check in the `render_markdown_to_body` method, this means that when the `_markdown` field is an empty string, it will get saved as such, but the `_body` field will remain the same as it previously was.

begedin:
Closes #336

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -68,7 +68,6 @@ class Project < ActiveRecord::Base
     end
 
     def render_markdown_to_body
-      return if long_description_markdown.blank?
       html = pipeline.call(long_description_markdown)
       self.long_description_body = html[:output].to_s
     end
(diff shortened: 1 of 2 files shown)


joshsmith:
I'm not sure if I agree with this. I implemented that check so that you couldn't get a blank body, which would be A Bad Thing™. Can we not just validate `markdown` when `present?`

begedin:
Then you can't create a project without an initial description, 

This seems like a more straightforward way to deal with it (as in, not deal at all). I can imagine where an admin or an owner would want to clear the description completely.

We could validate for presence if record is not new, though (on update, instead of on create).

joshsmith:
Closing as irrelevant because moving to Phoenix.

Issue #343, opened 2016-06-23
Enable coalesced find requests for GET requests on various endpoints


begedin:
What I think we need
- [x] Organization memberships - #344
- [x] Organizations - #345
- [x] Users - awaiting review in #347
- [x] User skills - awaiting review in #341
- [x] User categories
- [x] Skills
- [x] Comments
- [x] Posts

Others would make sense, but I think these are higher priority. Feel free to add to this description, though.

Also, I'm expecting duplicate code here, so might be a good idea to refactor into a concern once we have all the individual PRs merged.

PR #344, opened 2016-06-23
Enabled coalesced id filter on GET /organization/:id/memberships

begedin:
This is part of the solution for #343 

Also did some matcher cleanup along the way, since I thought i had an issue there.


joshsmith:
👍

PR #345, opened 2016-06-23
Enabled endpoint and coalesced id filter on GET /organizations

begedin:
Part of requirements for #343

We didn't have the route here at all. The reason I think we need the endpoint is we will likely have (or possibly already do) a scenario where wi get a user's organization memberships, and then want to get the organizations as well.

Since this reason is the only reason we need the endpoint, I made sure the id filter is required.


joshsmith:
👍

PR #349, opened 2016-06-24
Add more coalescing

### LINKED ISSUE #343: Enable coalesced find requests for GET requests on various endpoints

joshsmith:
Closes #343.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -15,29 +15,36 @@
 class CommentsController < ApplicationController
   before_action :doorkeeper_authorize!, only: [:create, :update]
 
-  def index
-    comments = Comment.where(post: params[:post_id]).includes(:user, :post)
+  # GET /posts/:number/comments
+  def post_index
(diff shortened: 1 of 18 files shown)


begedin:
@JoshSmith I see you commented out includes on `PostsController#show`. Is that permanent? I fixed the tests to reflect this.

joshsmith:
Updated and should be ready for review.

begedin:
Looks good to me, but I wrote part of it, so I'm not sure I'm enough for review.

joshsmith:
Good enough for now. If @christopherstyles wants to give thoughts after merge in case we need to fix something, we can open new issue then. Too time-sensitive, though.

Issue #351, opened 2016-06-24
Fix failing tests on first-run for someone without ENV


joshsmith:
After some consideration, here's a viable set of options to get us less ENV-dependent for development environments:
- [ ] Identify which integration tests we want to run fully to interact with services in CI (e.g. Facebook, Gravatar, Stripe, AWS) to be sure our interaction with those APIs works as expected
- [x] Skip certain tests if not in CI https://github.com/code-corps/code-corps-api/pull/153/files
- [x] Stub AWS response
- [ ] Mock Facebook calls in development
- [ ] Remove VCR altogether

We can do these in steps in separate PRs. Does not need to be tackled all at once.

joshsmith:
Here are the relevant tests that need skipped:

```
Pusher
spec/lib/code_corps/scenario/notify_pusher_of_comment_image_spec.rb
spec/lib/code_corps/scenario/notify_pusher_of_post_image_spec.rb

AWS
spec/models/comment_image_spec.rb
spec/models/organization_spec.rb
spec/models/post_image_spec.rb
spec/models/project_spec.rb
spec/models/user_spec.rb

Facebook
spec/requests/api/tokens_spec.rb
spec/requests/api/users_spec.rb
```

joshsmith:
Environment variables we rely on for each:

**AWS**
`AWS_ACCESS_KEY_ID`
`AWS_SECRET_ACCESS_KEY`
`CLOUDFRONT_DOMAIN`
`S3_BUCKET_NAME`

**Facebook**
`FACEBOOK_APP_ID`
`FACEBOOK_APP_SECRET`
`FACEBOOK_REDIRECT_URL`

**Pusher**
`PUSHER_APP_ID`
`PUSHER_FAKE`
`PUSHER_KEY`
`PUSHER_SECRET`

joshsmith:
Closing because moving to Phoenix.

PR #352, opened 2016-06-24
Stub AWS responses

christopherstyles:
To partially address #351, this change stubs AWS responses so that fake data is generated and no HTTP requests are made.

cc: @JoshSmith

### DIFF
--- a/spec/support/aws.rb
+++ b/spec/support/aws.rb
@@ -1,0 +1,1 @@
+Aws.config[:stub_responses] = true


joshsmith:
@christopherstyles so to clarify, should we explicitly disable this in places where we do want to test a response from AWS? Should we even do that?

christopherstyles:
@JoshSmith I don’t think we should be communicating with AWS at all from our tests. If we do need to test responses, we can specify [`stub_data`](http://docs.aws.amazon.com/sdkforruby/api/Aws/ClientStubs.html#stub_data-instance_method).

Issue #353, opened 2016-06-24
Allow post status to be changed


joshsmith:
We have a bug where you cannot open or close a post by setting its status. We need a failing test, and then make it pass.

joshsmith:
Closed by #358.

PR #358, opened 2016-06-24
#353 update posts_controller to allow status to be updated

rigelstpierre:


joshsmith:
👍  when the ENV works.

PR #359, opened 2016-06-25
Add a config file for Puma

christopherstyles:
This update moves configuration of Puma to an external file, providing finer-grained control of settings.

See the Heroku Dev Center [article](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#config) for additional info on recommended configuration options.


joshsmith:
👍

PR #360, opened 2016-06-25
Skip env tests

madisonsites:
@JoshSmith I wasn't sure how to write a test to skip tests or, more importantly, where I would put such a test, suggestions? 

Also, these two workers required the facebook env variables, so I added the `local_skip` to them as well: 

`spec/workers/add_facebook_friends_worker_spec.rb`
`spec/workers/add_facebook_profile_picture_worker_spec.rb`

This is my first time doing something like this for tests, so I'm extremely open to feedback on how this is structured. The tests _do pass_ for me now, so there's that.


joshsmith:
The fact that this is passing is already a good sign.

spec/requests/api/tokens_spec.rb
```diff
@@ -57,7 +57,7 @@
         end
       end
 
-      context "when facebook user does exist", vcr: { cassette_name: "requests/api/tokens/facebook_user_found" } do
+      context "when facebook user does exist", local_skip: true, vcr: { cassette_name: "requests/api/tokens/facebook_user_found" } do
```
joshsmith, line 60: This error you can probably ignore.

spec/models/organization_spec.rb
```diff
@@ -17,7 +17,7 @@
 
 require "rails_helper"
 
-describe Organization, :type => :model do
+describe Organization, :type => :model, local_skip: true do
```
joshsmith: just `type: :model`.

spec/rails_helper.rb
```diff
@@ -45,6 +45,11 @@
 end
 
 RSpec.configure do |config|
+  if Rails.env.test?
+    RSpec.configure do |c|
+      c.filter_run_excluding local_skip: true
+    end
+  end
```
joshsmith, line 52: I'm not sure if this is the way we want to approach it. This effectively says to never run the tests, because the `Rails.env.test?` will always be true when running tests.

I would recommend checking instead for an `ENV` variable that we would make available on CI, like `ENV["CI_BUILD_WITH_ENV"]` or something.

joshsmith, line 52: That would make this a one-line change to your PR. Assuming you're okay with that, and make that one Hound fix, then you can just squash your commits. Let me know if you're unfamiliar with that and I can show you how.

joshsmith:
@madisonsites 👍  can you rebase and squash?

joshsmith:
Closing in favor of #366. It's still your commit @madisonsites! I just forked _your_ repo and rebased off this repo so you could still have benefit of this commit.

Thanks!

PR #372, opened 2016-06-25
Add code theme preference for users

rigelstpierre:


joshsmith:
I believe this is missing a spec for the API and similarly allowing it to be set in the controller. You'll want to check the requests/users_spec for that.

PR #373, opened 2016-06-26
Speed up and backfill policy specs

christopherstyles:
Using `build_stubbed` significantly speeds up specs by reducing trips to the database. See the following:

https://robots.thoughtbot.com/use-factory-girls-build-stubbed-for-a-faster-test

Prior to this update, policy specs often ran in total around 25–28s. After the update, run times seem to be between 6–9s. :racehorse::dash: 

We can probably apply the same to other specs, quite easily for models.


joshsmith:
Rebased for you and will merge when ready.

PR #374, opened 2016-06-26
CSV skills import

### LINKED ISSUE #320: Build an import pipeline for adding skills from CSV
We need a way to parse a CSV of skills and upload them. This ideally should be idempotent and can be used as a way for developers to seed the app, as well as seed our own production database.

schneidmaster:
Closes #320

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ vendor/bundle/*
 dump.rdb
 
 coverage
+
+/orgnizations
+/projects
+/users
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,6 @@ dump.rdb
 
 coverage
 
-/orgnizations
+/organizations
 /projects
 /users
--- a/app/policies/import_policy.rb
+++ b/app/policies/import_policy.rb
@@ -14,6 +14,6 @@ class ImportPolicy
     return false unless user.admin?
 
     # Can create import.
-    return true
+    true
   end
 end
--- a/app/policies/import_policy.rb
+++ b/app/policies/import_policy.rb
@@ -1,0 +1,19 @@
+class ImportPolicy
+  attr_reader :user, :import
+
+  def initialize(user, import)
+    @user = user
+    @import = import
+  end
+
+  def create?
+    # Cannot create if there's no user
+    return false unless user.present?
+
+    # Cannot create if they're not an admin.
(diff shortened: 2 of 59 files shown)


.gitignore
```diff
@@ -32,4 +32,8 @@ vendor/bundle/*
 
 dump.rdb
 
-coverage
\ No newline at end of file
+coverage
+
+/orgnizations
+/projects
+/users
```
joshsmith, line 39: looks like it should be `/organizations`

schneidmaster, line 39: it should be, but it's misspelled in the model too, so 😛

joshsmith, line 39: Well. 😢 

Can we change in both then? I'll update in S3.

schneidmaster, line 39: Yeah will do

joshsmith:
I renamed organizations folder in all S3 so we're good there.

app/policies/import_policy.rb
```diff
@@ -0,0 +1,19 @@
...
+  def create?
+    # Cannot create if there's no user
+    return false unless user.present?
+
+    # Cannot create if they're not an admin.
+    return false unless user.admin?
+
+    # Can create import.
+    true
+  end
```
begedin, line 18: I would just go with `return true if user.admin?` here. Sure, it does't follow our usual convention, but it's much more readable, I think.

begedin:
@JoshSmith @schneidmaster 

I added 2 commits here

The first one is relatively simple. It adds a slug to the skill model, so we can, in the future, implement a skill page. Unlike users and organizations, the skills slug won't be accessed from the common slugged route (I assume) and will instead be on "skills/:slug" or something like that. Due to that, the behavior is not exactly the same.

The second commit adds handling of import failures. For each row that fails, it's status is set to 3 (`:failed`) and an `ImportSkillFailure` record is created, which has 4 columns:
- `skill` - the skill that was failed to update. it we failed at creating a new skill, this will be null
- `import` - the import record that failed
- `issues` - `skill.errors.full_messages`, a text array column
- `data`\- the csv row that failed to import. JSON column

All of these work just fine with our serializer.

I also added a very simple endpoint to fetch a list of failures, which can be enhanced as we need it.

schneidmaster:
LGTM 👍

PR #392, opened 2016-07-05
Split name into first and last, rewrite model and API specs

joshsmith:
This splits the name into first and last name.


begedin:
Looks good, but we should probably hold off on merging it until ember-side is ready.

PR #230, opened 2016-02-13
Update paperclip and shoulda matchers

joshsmith:
Not working yet. Could use assistance to get these specs passing. Not sure what the issues are, but they lie with changes in shoulda matchers.

PR #394, opened 2016-07-05
Rails upgrade from 4.2.5.1 to 5.0.0

rigelstpierre:


joshsmith:
👍  looks good.

I think we need to open up a separate issue to track the Clearance issue.

Do you know if this resolves #230 @rigelstpierre?

PR #431, opened 2016-07-07
Add docker

joshsmith:
This adds Docker to replace Vagrant, since that's so yesterday. Literally.

I still need to rewrite docs to account for this, but it appears to be working. Pushing up so we can test locally some.


bin/setup
```diff
@@ -3,41 +3,37 @@
 # Exit if any subcommand fails
 set -e
 
+# Import colors
+. colors
```
npendery: Got an error with this. I changed it to be `. bin/colors` and it worked

bin/setup
```diff
@@ -3,41 +3,37 @@
 # Exit if any subcommand fails
 set -e
 
+# Import colors
+. bi/colors
```
christopherstyles: Missing an “n” here I believe (`. bin/colors`).

joshsmith: Fixed.

Guardfile
```diff
@@ -0,0 +1,70 @@
...
+# Note: The cmd option is now required due to the increasing number of ways
+#       rspec may be run, below are examples of the most common uses.
+#  * bundler: 'bundle exec rspec'
+#  * bundler binstubs: 'bin/rspec'
+#  * spring: 'bin/rspec' (This will use spring if running and you have
+#                          installed the spring binstubs per the docs)
+#  * zeus: 'zeus rspec' (requires the server to be started separately)
+#  * 'just' rspec: 'rspec'
+
+guard :rspec, cmd: "bundle exec rspec" do
```
christopherstyles, line 27: I’d suggest adding a spring binstub for rspec with `bundle exec spring binstub rspec` then changing this line to:

``` ruby
guard :rspec, cmd: "bin/rspec" do
```

In my tests it was trimming 7–9 seconds off load time.

christopherstyles:
👍 :shipit:

PR #442, opened 2016-07-12
Add API Blueprint

joshsmith:
This adds the beginning of API Blueprint documentation.

We should still:
- [ ] Add the ability to push to Apiary in Circle
- [x] Add README notes for generating (using Docker) and viewing (in your Browser, via aglio) API docs
- [x] Add some empty areas for future endpoints


blueprint/api.apib
```diff
@@ -0,0 +1,146 @@
...
++ Response 200 (application/json)
+    + Attributes (OAuth Valid Response)
+
+# Group Users
+
+## User [/users/{id}]
+
++ Parameters
+
+  + id: 1 (string)
```
npendery, line 39: I think after the 1 this should be: `(required, number) - ID of the User in form of an integer`

blueprint/api.apib
```diff
@@ -0,0 +1,146 @@
...
+
+            Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+
++ Response 200 (application/json)
+    + Attributes (OAuth Valid Response)
+
+# Group Users
+
+## User [/users/{id}]
+
```
npendery, line 36: Should we add an attributes section?

joshsmith, line 36: Should be covered by the `Attributes (User Response)`.

joshsmith:
Will make more changes once this is up.

PR #444, opened 2016-07-12
Categories data added to API documentation

### LINKED ISSUE #400: Document Categories API

npendery:
Closes #400

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -37,6 +37,82 @@ Exchange credentials for `access_token`.
 
 # Group Categories
 
+## Category [/categories/new]
+
+### Create a category [POST]
+
++ Request
+  + Attributes (OAuth Grant Request)
+
++ Headers
+
+  Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+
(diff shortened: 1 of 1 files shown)


api-docs/api.apib
```diff
@@ -33,6 +33,40 @@ Exchange credentials for `access_token`.
 
 # Group Categories
 
+## Categories [/categories]
+
+### Retrieve categories index [GET]
```
joshsmith: I think this should maybe rewritten as:

> List all categories

joshsmith:
I think ordering should be:
- Create
- Retrieve
- Update
- Delete
- List

Taking Stripe's docs as our guide.

What do you think?

joshsmith:
@npendery I also forgot to remove the `api-docs` directory in my PR. We should be using `blueprint/api.apib` isntead.

npendery:
@JoshSmith I feel List would be first since I tend to think of that as the base of the other operations, but not a big deal. And sounds good about the location.

Will add that order and the location to the journey issue info.

joshsmith:
This looks good to me. Do you want to rebase and squash @npendery and then you can merge it in yourself.

PR #446, opened 2016-07-13
Remove custom install documentation

joshsmith:
I think the custom install documentation will lead to a lot of heartache when people should ideally just use Docker.

Any votes against this?


christopherstyles:
👍

npendery:
I think it makes sense to remove. Having the same setup will allow people to more easily collaborate by limiting the chance of their environment being different.

PR #451, opened 2016-07-15
Improve Apiary documentation process

joshsmith:
This PR:
- adds commands to push the API docs to Circle
- adds the ability to run an apiary server locally with Docker
- adds more to the documentation itself


begedin:
👍

PR #452, opened 2016-07-15
Fix some minor issues with squashing docs

joshsmith:
Fixes issues raised in Ember PR.


christopherstyles:
Should we link out to the docs on squashing commits, for more in-depth instruction? It’s explained pretty well there: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

Otherwise 👍

joshsmith:
@christopherstyles Yeah, I do so in `CONTRIBUTING.md`, which is the entrypoint to these docs.

PR #456, opened 2016-07-19
Add Comment Image API docs

thomascchen:
This is a first attempt at adding API documentation for Comment Images. 
Never used API Blueprint before 😬  so hope this isn't wildly off the mark.

Closes https://github.com/code-corps/code-corps-api/issues/401

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -115,6 +115,25 @@ Exchange credentials for `access_token`.
 
 # Group Comment Images
 
+## Comment Image [/comment_images]
+
+### Create a Comment Image [POST]
+
++ Request
+  + Attributes (OAuth Grant Request)
+
++ Headers
+
+  Authorization: Bearer 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+
++ Response 201 (application/vnd.api+json)
+
+  + Attributes (Comment Image Response)
+
++ Response 422 (application/vnd.api+json)
+
+  + Attributes (Unprocessable Entity Response)
+
 # Group Comment User Mentions
 
 # Group Comments
@@ -248,6 +267,28 @@ Exchange credentials for `access_token`.
 ## Categories Response (object)
 + categories
   + data(array[Category], required)
+
+## Comment Image (object)
++ filename: `file name` (string)
+
+## Comment Image Relationships Base (object)
++ comment
+  + data(Comments Relationship)
++ user
+  + data(Users Relationship)
+
+## Comment Image Response Base (object)
++ id: `1` (string)
++ type: `comment_images` (string)
++ attributes(Comment Image)
++ relationships(Comment Image Relationships Base)
+
+## Comment Image Response (object)
++ data(Comment Image Response Base)
+
+## Comments Relationship (object)
++ id: 1 (string)
++ type: comments (string)
 
 ## Creation Conflict Response (object)
 + title: `Conflict` (string)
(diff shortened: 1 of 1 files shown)


blueprint/api.apib
```diff
@@ -115,6 +115,18 @@ Exchange credentials for `access_token`.
 
 # Group Comment Images
 
+## Comment Image [/comment_images]
+
+### Create a Comment Image [POST]
++ Request
```
npendery, line 122: Just a styling thing here, break out the lines with double spacing

blueprint/api.apib
```diff
@@ -115,6 +115,18 @@ Exchange credentials for `access_token`.
 
 # Group Comment Images
 
+## Comment Image [/comment_images]
```
npendery, line 118: This should be `## Comment Image [/comment_images/new]` for the Post

blueprint/api.apib
```diff
@@ -225,6 +237,28 @@ Exchange credentials for `access_token`.
 + categories
   + data(array[Category], required)
 
+## Comment Image (object)
++ filename: `file name` (string)
+
+## Comment Image Relationships Base (object)
```
npendery, line 274: Shouldn't both the relationships below just be a single object and not an array? ie:

```
+ comment
  + data(Comments Relationship)
+ user
  + data(Users Relationship)
```

blueprint/api.apib
thomascchen, line 118: I'm fairly sure that `[/comment_images]` is the correct path for the `create` action. See http://guides.rubyonrails.org/routing.html#crud-verbs-and-actions

green-arrow, line 118: If we're following JSON API, then the correct endpoint is `POST /comment_images`

blueprint/api.apib
thomascchen, line 122: I can make this change, but our spacing seems to be a little inconsistent here. Are there conventions or a style guide for API Blueprint stuff? Or should we identify/create some?

green-arrow, line 122: It definitely is inconsistent. In traditional markdown, things that are separated by a line break in the document need a line in between them in the `.md` file. I would vote for the double spacing.

thomascchen:
Thanks for the comments! 😄  I've fixed the relationship and adjusted the spacing per @npendery and @green-arrow's suggestion, but I'd like to open a separate issue for identifying or creating some set of conventions for our API docs, for things like spacing. I'd prefer to keep this PR focused on accurately describing the API for comment image.

npendery:
LGTM! 👍

blueprint/api.apib
```diff
@@ -115,6 +115,25 @@ Exchange credentials for `access_token`.
...
+## Comment Image [/comment_images]
+
+### Create a Comment Image [POST]
+
++ Request
+  + Attributes (OAuth Grant Request)
+
++ Headers
+
+  Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
```
christopherstyles: This should be `Bearer` instead of `Basic`. Been meaning to open a separate PR to correct it in existing examples. 

Also, do you have any thoughts about potentially replacing the `5e884898da28…` string with maybe a token indicator? So it would look something like: `Authorization: Bearer {access_token}`. I didn’t see a precedent for this in the Blueprint docs, but I think the meaning comes across, and is a little easier to mentally parse than trying to figure out where this string came from.

green-arrow: I agree with replacing it so it becomes `Authorization: Bearer <access_token`. I think that might be better served in its own issue / PR though, just for clarity.

joshsmith:
@thomascchen are you planning more changes here? What are the next steps?

thomascchen:
@JoshSmith just wanted to make the change suggested @christopherstyles and @green-arrow had suggested, which I've done now. Also squashed and rebased so this is ready for review again. 👍

npendery:
LGTM

PR #458, opened 2016-07-20
API style guide specs

npendery:
This is to start the discussion on how we want to style the API documentation for contributors


joshsmith:
This looks good now I think but doesn't seem to be actionable at the moment. What's the next step here?

thomascchen:
@npendery can we close this in favor of discussion in https://github.com/code-corps/code-corps-api/issues/463? If there are specific suggestions/recommendations that you'd like to not get lost here, can you bring them into that issue?

npendery:
Yup

PR #460, opened 2016-07-21
Added documentation for the comment user mentions API

### LINKED ISSUE #402: Document Comment User Mentions API

green-arrow:
Resolves #402

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -16,18 +16,20 @@ The staging API can be found at `https://api.pbqrpbecf.org`.
 The remote development API can be found at `https://api.pbqrpbecf-qri.org`.
 
 # Authentication
-When using the API's authenticated endpoints you can authenticate users by exchanging the `username` and `password` for an OAuth token.
+
+When using the API's authenticated endpoints you can authenticate users by exchanging the `username` (email address) and `password` for an OAuth token.
 
 ## OAuth Token Exchange [/oauth/token]
-Exchange credentials for `access_token`.
 
-### Exchange credentials for token [POST]
+Exchange email and password for `access_token`.
+
+### Exchange email and password for token [POST]
+
 + Request
     + Attributes (OAuth Grant Request)
     + Headers
 
-            Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
-            Content-Type: application/x-www-form-urlencoded
+            Content-Type: application/vnd.api+json
 
 + Response 200 (application/json)
     + Attributes (OAuth Valid Response)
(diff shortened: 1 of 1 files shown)


blueprint/api.apib
```diff
@@ -232,6 +251,30 @@ Exchange credentials for `access_token`.
 
 + status: 409 (number) - HTTP status code
 
+## Comment Relationship (object)
++ id: 1 (string)
++ type: comments (string)
+
+## Comment User Mention (object)
```
npendery: We should split this into `Comment User Mention (Object)` with just the attributes and then a `Comment User Mention Response Base (Object)` that includes the attribute group as `+ attributes(Comment User Mention, required)`
(At least that is how we have been doing it)

green-arrow: So I was actually going to ask about this. In the categories section, we're doing that, but it seems like it's misused. There is a `Category` object that is used directly in the `Categories Response` in the data array. This is incorrect, since resource must always be defined as their id, type, and attributes according to JSON API. I'm ok with making this a separate object, but don't see where we'd get any reuse out of it.

npendery: Reading from [this](http://jsonapi.org/format/#document-resource-objects) it seems like we should not have a `Response Base`, but have the `main object` (Comment User Mention) and another `attributes object` (Comment User Mention Attributes) that is included in the main object  that details out the non-id attributes?

green-arrow: That would definitely map to the spec a lot better. I'm just wondering if it's even necessary to separate out the attributes object since that object will never be reused anywhere else. The whole resource has to be present, not just `attributes`.

npendery: Sounds good to me! Makes sense to deconstruct the attributes but not the relationships because those are reused

christopherstyles: How I’ve been doing relationships is to break them into a base/attributes/collection group of 3 objects that are included from a parent response:

``` markdown
<!--- the parent response --->
## User Response (object)
+ data
  + id: `1` (string)
  + type: `users` (string)
  + attributes(User, required)
  + relationships(User Relationships Base, required)

<!--- our base relationships collection --->
## User Relationships Base (object)
+ include User Skills Relationships

<!--- attributes for a user skill. notice the singular name -->
## User Skills Relationship (object) 
+ id: 1 (string)
+ type: user_skills (string)

<!--- a collection of user skill relationships. notice the plural name. This is what we include in the base response. -->
## User Skills Relationships (object) 
+ user_skills
  + data(array[User Skills Relationship], required)
```

This would support a json object that looks like:

``` json
{
  "data": {
    "id": "1",
    "type": "users",
    "attributes": {
      "created_at": "2016-07-08T03:03:51.967Z",
      "email": null,
      "first_name": "Madaline",
      "last_name": "Conroy",
      "other…": "attributes…"
    },
    "relationships": {
      "user_skills": {
        "data": [
          {
            "id": "1",
            "type": "user_skills"
          }
        ]
      }
    }
  }
}
```

Seems a bit verbose, but is helpful if the relationships are being included in multiple parent responses.

green-arrow: I like this method of splitting things out. 👍 

Edit: @christopherstyles the only thing I would suggest is splitting out the `User Response (object)` into a `User Response` and `User`, that way we could reuse the `User` definition in the response for multiple users via `data(array[User])`. Thoughts?

green-arrow: Quick question @christopherstyles - what do you see as the benefit for splitting out the relationship object into a singular and plural? Would both the singular and plural objects be reused elsewhere? And if so, where would each be used?

joshsmith:
Someone needs to take this on and update according to the RFC in #463.

joshsmith:
Made some changes and updated per RFC (I think). And also wrote some in-plain-English description for the endpoint.

npendery:
What the eff happened here? ^^

joshsmith:
👍 but could use someone who can come in and squash now.

PR #461, opened 2016-07-22
Added documentation for comments API

### LINKED ISSUE #403: Document Comments API

begedin:
Closes #403

I did a sort of risky thing and decided to try a different naming approach as a sort of proposal to how we could do it in the future.

My issue with what we currently have is that the object names are very similar.

For example, with the current scheme, we would have:
- Comments Relationships - shows related comments in other objects, ex. posts, users, etc.
- Comments Relationship - a single { id, type } hash, multiple of which are contained in Comments Relationships
- Comment Relationships - a list of objects related to the current comment, part of a single Comment's response. With my current proposal, we would have. In this specific case, it would include Users Relationships, Posts Relationships and Comment User Mentions Relationships

With my proposal, we would have

```
Comment
+ body, markdown, created_at, etc.

Head For Comment
+ id, type

Attributes For Comment
+ attributes: (Comment, required)

Relationships For Comment
+ relationships: 
   + include Related Posts
   + include Related Users
   + Related Comment User Mentions

Comment Response
+ Head For Comment
+ Attributes For Comment
+ Relationships For Comment
```

If another object is related to comments, it would have a (with an example of a post)

```
Relationships For Post
+ relationships:
   + include Related Comments
```

Where related comments would look like:

```
Related Comments
  + data(array[Head For Comment])
```

In my opinion, I think it makes the documentation less confusing, but I'm always open to critique. @JoshSmith @green-arrow @npendery, what do you think?

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -138,6 +138,148 @@ Exchange credentials for `access_token`.
 
 # Group Comments
 
+## Comments [/comments]
+
+<<<<<<< HEAD
+### Filter comments by list of ids [GET /comments{?filter[id]}]
+
++ Parameters
+
+    + `filter[id]`: `1,2,3` (string, required) - Comma separated string of `ids` to filter by.
+
+        Technically, the parameter is `filter: { id: '1,2,3' }`, which ends up being serialized into `filter[id]`. However, synthax constraints to not allow us to define it in the documentation this way.
+
+
++ Request
+
+    + Headers
+
+        Accept: application/vnd.api+json
+
+        Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+    + Attributes (Comments Response)
+
+## List comments for a specified post [GET /posts/{number}/comments]
+
++ Parameters
+
+    + number: 1 (number, required) - Post number, uniquely identifying a post
+
++ Request
+
+    + Headers
+
+        Accept: application/vnd.api+json
+
+        Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+    + Attributes (Comments Response)
+
+### Create a comment [POST]
+
++ Request
+    + Headers
+
+        Accept: application/vnd.api+json
+
+        Authorization: Bearer {access_token}
+
+    + Attributes (Comment Create Request)
+=======
+### Create a comment [POST]
+
++ Request
+
+    + Attributes (Comment Create Request)
+
++ Headers
+
+    Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+
++ Response 201 (application/vnd.api+json)
+
+    + Attributes (Comment Response)
+
++ Response 422 (application/vnd.api+json)
+
+    + Attributes (Unprocessable Entity Response)
+
+### Update a comment [PATCH /comments/{id}]
+
++ Request
+
+    + Attributes (Comment Update Request)
+
++ Headers
+
+    Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+>>>>>>> dcb1458... Added documentation for comment API
+
++ Response 201 (application/vnd.api+json)
+
+    + Attributes (Comment Response)
+
++ Response 422 (application/vnd.api+json)
+
+    + Attributes (Unprocessable Entity Response)
+
+<<<<<<< HEAD
+## Comment [/comments/{id}]
+
++ Parameters
+
+    + id: `1` (string, required) - Comment id, uniquely identifying a comment
+
+### Update a comment [PATCH]
+
++ Request
+
+    + Headers
+
+        Accept: application/vnd.api+json
+
+        Authorization: Bearer {access_token}
+
+    + Attributes (Comment Update Request)
+
++ Response 201 (application/vnd.api+json)
+
+    + Attributes (Comment Response)
+
++ Response 422 (application/vnd.api+json)
+
+    + Attributes (Unprocessable Entity Response)
+=======
+### List comments with specified ids [GET]
+
++ Request
+
+    + Attributes (Id Filter)
+
++ Headers
+
+    Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+
++ Response 200 (application/vnd.api+json)
+
+    + Attributes (Comments Response)
+
+## List comments for a specified post [GET /posts/:number/comments]
+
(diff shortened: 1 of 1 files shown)


green-arrow:
@begedin I think we definitely need to arrive at a standard for how we create the data structures. In my opinion I think we should consider trying to mirror the naming of the JSON API spec for consistency. I would suggest something like this:

```
Comment Response (object)
+ data(Comment Resource)

Comments Response (object)
+ data(array[Comment Resource])

Comment Resource (object)
+ include Comment Resource Identifier
+ attributes
  + body, markdown, created_at, etc
+ relationships
  + post
    + data(array[Post Resource Identifier])
    + links

Comment Resource Identifier (object)
+ id, type

Post Resource Identifier (object)
+ id, type
```

The reason I didn't break out the `attributes` and `relationships` section is because those two items are unique to this resource, and can't be reused across other data structures.

Thoughts?

joshsmith:
@green-arrow I really like the approach of using JSON API spec for the naming. That's what we're going for anyway, right?

@begedin what do you think about this? You got the conversation started here, but definitely like the direction of JSON API naming conventions instead.

thomascchen:
@JoshSmith @green-arrow @begedin What do you think about opening a new issue where we can combine this discussion and https://github.com/code-corps/code-corps-api/pull/458 into a discussion of standards/conventions for API blueprint? I've been meaning to open an issue and do some poking around to see if anyone else has come up with some good conventions we could follow. Just haven't had a minute...

green-arrow:
@thomascchen I'm good with opening up an issue and detailing the requirements there. Should we add an additional 'guide' in the README or somewhere else that lays out those details? If so, maybe we should just open a PR with that new guide and iterate on that?

joshsmith:
@thomascchen sure, that's fine. Also okay having conversation here, but it may be a better to break it out separately. I'd definitely like to nail this down ASAP since I'll be trying to spend time myself to help wrap this whole project up quickly and move onto Phoenix-land.

joshsmith:
@green-arrow if we're going to open a new issue, I think maybe keep it as an RFC? Then we can move onto a PR that documents.

green-arrow:
@JoshSmith that sounds good to me

green-arrow:
If you'd like @JoshSmith, I can open up the issue now and copy what I had above to get things started.

joshsmith:
@green-arrow that'd be great.

begedin:
Sorry for delaying on this, everyone. I've been busy on another project. I'll be able to take this on tomorrow morning and wrap it up.

blueprint/.editorconfig
```diff
@@ -0,0 +1,2 @@
+[*]
+indent_size = 4
```
begedin: Considering what @christopherstyles said in https://github.com/code-corps/code-corps-api/issues/463#issuecomment-235750923, I took the liberty of adding an `.editorconfig` for our blueprint folder, which overrides the root `.editorconfig`

begedin:
I did some more work here, so this should be good to go now. I used the example of User Categories and our discussion in #463 to make changes here, so it should be up to our standards.

I had to add an `Id Filter` object, which will, I expect, be reused in some of our other endpoints. Wasn't sure where to place it, since the structure is not model specific, so I added it to the end of the file for now. We can reorder everything later.

I also added an indentation rule for our `.editorconfig`, based on what @christopherstyles said in https://github.com/code-corps/code-corps-api/issues/463#issuecomment-235750923

christopherstyles:
Great idea on the `.editorconfig` @begedin, thanks!

blueprint/api.apib
```diff
@@ -138,6 +138,68 @@ Exchange credentials for `access_token`.
...
+
++ Response 201 (application/vnd.api+json)
+
+    + Attributes (Comment Response)
+
++ Response 422 (application/vnd.api+json)
+
+    + Attributes (Unprocessable Entity Response)
+
+### List comments with specified ids [GET]
```
npendery, line 258: In the specs we have that this is required to put an id filter in so we should add a `400` response. I also was looking at the apib docs and we should add the url ext such as `/comments{?id_filter}` or similar.

blueprint/api.apib
```diff
@@ -138,6 +138,68 @@ Exchange credentials for `access_token`.
...
+
++ Headers
+
+    Authorization: Basic 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+
++ Response 201 (application/vnd.api+json)
+
+    + Attributes (Comment Response)
+
++ Response 422 (application/vnd.api+json)
```
npendery, line 227: Specs also have a `401` error for Not Authorized so that should be added as well

blueprint/api.apib
```diff
@@ -138,6 +138,68 @@ Exchange credentials for `access_token`.
...
```
npendery, line 208: Also should have a `401` response

blueprint/api.apib
```diff
@@ -138,6 +138,68 @@ Exchange credentials for `access_token`.
...
+
+    + Attributes (Unprocessable Entity Response)
+
+### List comments with specified ids [GET]
+
++ Request
+
+    + Attributes (Id Filter)
```
npendery, line 262: Nice! Definitely will be reused so good call as an object

begedin, line 262: @npendery, I'm afraid this is wrong, so I'll have to fix it. It needs to go under parameters and we can't use an attributes call there. Possibly, we can just include it as an object, so I'll try that.

blueprint/api.apib
```diff
@@ -138,6 +138,93 @@ Exchange credentials for `access_token`.
...
 
+## Comments [/comments]
+
+### Filter comments by list of ids [GET /comments{?filter[id]}]
+
++ Parameters
+
+    + `filter[id]`: `1,2,3` (string, required) - Comma separated string of `ids` to filter by.
+
+        Technically, the parameter is `filter: { id: '1,2,3' }`, which ends up being serialized into `filter[id]`. However, synthax constraints to not allow us to define it in the documentation this way.
```
begedin, line 150: This is the last thing that bugs me in the documentation at the moment. JSON API (and ember-data) coalesce ids by applying the following hash to the query string:

``` Javascript
{
  filter: {
    id: "1,2,3,etc"
  }
}
```

This then ends up serialized in a get request as `?filter[id]=1,2,3`. Thing is, it looks like API Blueprint doesn't support that. Giving a hash as a parameter fails completely, as does an object. Providing `filter[id]` as a single parameter, as done here, sort of works, but still not really, since it doesn't apply it to the test request we can do on the documentation page. I'm not sure if there's a better way to do this, so I'm open for suggestions.

PR #465, opened 2016-07-26
Document Skills API

### LINKED ISSUE #424: Document Skills API

npendery:
This update adds API documentation for Skills. The following actions have been documented:
- Create a skill – `POST /skills`
- Find a skill – `GET /skills/{id}`
- List/search skills - `GET /skills`

Closes #424

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -382,6 +382,75 @@ This allows A Skill to be under different Roles and a Role to have multiple Skil
 
 # Group Skills
 
+This endpoint retrieves skills that can be applied to both users (through user_skills) and roles (through role_skills)
+
+## Skills [/skills?{filter}]
+
+### List all skills or filter by parameter [GET]
+
++ Parameters
+
+    + filter: `1` (enum[number], optional) - Id query to filter returned skills
+
+    + query: `elixir` (string, optional) - Text query to filter returned skills
+
++ Request
+
+    + Headers
+
+        Accept: application/vnd.api+json
+
++ Response 200 (application/vnd.api+json)
+
+    + Attributes (Skills Response)
+
+### Create a skill [POST]
+
++ Request
+
+    + Headers
+
+        Accept: application/vnd.api+json
+        Authorization: Bearer <access_token>
+
++ Response 200 (application/vnd.api+json)
+
+    + Attributes (Skill Response)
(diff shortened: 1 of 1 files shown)


joshsmith:
@npendery made some changes here. Some of that affects categories and not just skills.

A couple things I think are missing here:
- filtering in the endpoint
- searching for skills (the autocomplete part)

Both are buried in the `index` action.

We can tackle those separately if we add some issues for them.

npendery:
@JoshSmith just wondering why you decided to use this PR to change category items? Wouldn't we want to separate that out in a separate PR?

joshsmith:
@npendery was just for simplicity's sake. I normally would want to do that but was having trouble organizing thoughts around your PR without the organizational work there.

blueprint/api.apib
```diff
@@ -595,6 +662,26 @@ This allows A Skill to be under different Roles and a Role to have multiple Skil
...
++ data(Skill Resource)
+
+## Skills Response (object)
++ data(array[Skill Resource])
+
+## Skill Resource (object)
++ include Skill Resource Identifier
++ attributes
+  + title: `Ruby`
+  + description: `Knowledgeable about the Ruby Programming Language`
```
JoshSmith: Why'd you change this from the Elixir implementation I had pushed up? Thought the description there was clearer (and would be more fun for our new Elixir folks).

joshsmith:
Did we lose a number of changes here? I feel like there's a lot missing in what I had made and can't tell since it was rebased.

joshsmith:
Yeah, this lost all the very real changes I was trying to make to suggest some rework of grouping of the resources. Not sure how to proceed.

npendery:
Oh crap. Not sure how that happened....

joshsmith:
@npendery is this good to review again? If it was, then I approve.

joshsmith:
I think the `query` and the `filter` should be broken out into two separate use cases, or (and maybe better) we should spend some time defining what each does in plain English a little better.

I can tackle that if you'd like, just assign to me.

npendery:
@JoshSmith yeah, wasn't sure how we would want to display that. Will update and separate them out.

rigelstpierre:
🐑 🇮🇹

PR #469, opened 2016-07-27
Use “apiaryio” gem over “apiary”

christopherstyles:
Oops! The intended gem for the [Apiary.io client](https://github.com/apiaryio/apiary-client) is `apiaryio` instead of `apiary`. This update swaps out the correct gem.


joshsmith:
👍

PR #464, opened 2016-07-23
Document Users API

christopherstyles:
This update adds API documentation for Users. The following actions have been documented:
- Create a user – `POST /users`
- Find a user – `GET /users/{id}`
- Check email availability – `GET /users/email_available?email={email}`
- Check username availability – `GET
  /users/username_available?username={username}`
- Update a user – `PATCH /users/{id}`
- Update “me” – `PATCH /users/me`
- List users – `GET /users?filter[id]=1,2`
- Show authenticated user – `GET /user`

Resolves #390, closes #455.

cc: @JoshSmith

PR #471, opened 2016-07-27
Fix role skill resource relationships

christopherstyles:
In #464 these data structures were renamed as resource identifiers. This update reflects the change within the role skills relationship collections.


npendery:
Sweet thanks for the update, LGTM!

christopherstyles:
Thanks @npendery!

PR #478, opened 2016-07-28
Add documentation for projects API endpoint

### LINKED ISSUE #421: Document Projects API

begedin:
Closes #421, but there are a couple of items here that are worth discussing first
- We have an endpoint to list all projects from a specified slugged route. This is a remnant from when a project could belong to a user or an organization. It can only belong to an organization now. Should it be removed, or possibly redefined to work differently? Perhaps the `GET /projects` endpoint should support an `organization_id` parameter?
- We have an endpoint to fetch a project by `:slugged_route/:slug`, which is redundant, really. Our client will always have access to the project id when needing to fetch a project and we already have an endpoint to fetch by id directly. Should we remove the fetch by slug endpoint?
- Project#show includes github_repositories in the payload (not just the relationships. the actual includes as well). Should that be removed, considering it doesn't really do anything right now? Should we even include the github relationship in the payload, considering we don't have it working yet?
- Project#create and Project#update currently allow us to send a slug alongside the request. Should it not be auto-generated from the title?
- Project#create currently sets organization as part of the attributes payload, which is not according to JSON API spec. I have added the relationship, even though it doesn't match our current API. We cannot infer it from the current user, since the current user can belong to multiple organizations where they are able to create projects.
- Should we have a Project#create at all right now?

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -689,6 +689,76 @@ Write privileges are for admin users only, read can be accessable by any user.
 
 # Group Projects
 
+## Projects [/projects]
+
+### List all projects [GET]
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Projects Response)
+
+### Retrieve a project [GET /projects/:id]
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Project Response)
+
+### Retrieve all projects for organization specified by slug [GET /:slug/projects]
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Projects Response)
+
+### Retrieve a project by slug [GET /:slug/:project_slug]
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Project Response)
+
+### Create a project [POST]
+
++ Request
+
+  + Attributes (Project Create Request)
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Project Response)
+
+### Update a project [PATCH /projects/:id]
+
++ Request
+
+  + Attributes (Project Update Request)
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Project Response)
+
 # Group Roles
 
 This endpoint is for creating and returning the roles available for users.
@@ -1477,6 +1547,10 @@ This endpoint retrieves sets of users and skills to act as a join table between 
 + filter:
     + id: `1,2,3,...` (string, required) - A list of ids, in a single string, comma-separated
 
+## Github Repository Resource Identifier
++ id: `1` (string)
++ type: `github_repositories` (string)
+
 ## No Content Response (object)
 
     {}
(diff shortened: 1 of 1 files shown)


joshsmith:
If someone hits `:slugged_route/:slug` directly in the client, I don't believe they will have immediate access to the project `id`.

joshsmith:
Remove the github repositories from the documentation and we won't add it in Phoenix.

joshsmith:
You should be able to modify the project slug at will (though with a warning when changing it).

joshsmith:
Sounds good on fixing `organization` in `attributes`. Even though it works incorrectly now, we will want it to work correctly later.

We should an issue in Ember project that's labeled `blocked`, and one in Phoenix project to make sure we catch it specifically.

joshsmith:
`Project#create` can be punted some, but we might as well document it. Our API documentation is _way_ ahead of our API on the Phoenix side, anyway. Maybe we'll just add notes that they are pending.

joshsmith:
I think later a project _will_ possible to belong to a user again, so I think that should remain as flexible as possible for forward compatibility.

joshsmith:
The language of "Find" here differs from our "Retrieve". Which one are we settling on? I prefer "Retrieve" to stay with CRUD.

blueprint/api.apib
```diff
@@ -298,6 +298,76 @@ Write privileges are for admin users only, read can be accessable by any user.
...
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Projects Response)
+
+### Find a project [GET /projects/:id]
```
joshsmith: Find => Retrieve.

blueprint/api.apib
```diff
@@ -298,6 +298,76 @@ Write privileges are for admin users only, read can be accessable by any user.
...
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Project Response)
+
+### Find all projects for organization specified by slug [GET /:slug/projects]
```
joshsmith: Find => Retrieve.

blueprint/api.apib
```diff
@@ -298,6 +298,76 @@ Write privileges are for admin users only, read can be accessable by any user.
...
+
++ Headers
+
+  Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+  + Attributes (Projects Response)
+
+### Find project by slug [GET /:slug/:project_slug]
```
joshsmith: Find => Retrieve.

blueprint/api.apib
```diff
@@ -842,10 +916,63 @@ When creating a user category, the authenticated user is assumed by the server s
...
+        + data(array[Skill Resource Identifier])
+    + organizations
+        + data(Organization Resource Identifier)
+
+## Project Resource Identifier (object)
 + id: `1` (string)
 + type: `projects` (string)
 
+## Project Attributes (object)
++ slug: 'example_slug' (string) - Case insensitive unique string identifier for the project
```
joshsmith: I would use `example-slug` since I think it parameterizes by default.

blueprint/api.apib
```diff
@@ -842,10 +916,63 @@ When creating a user category, the authenticated user is assumed by the server s
...
++ title: 'An example project' (string) - A title for the project
++ description: 'An example project description' (string) - A short plain text description for the project
++ icon_thumb_url: 'example.com/icon_thumb.png' (string) - A url pointing to the project thumbnail
++ icon_large_url: 'example.com/icon_large.png' (string) - A url pointing to a large icon for the project
++ long_description_body: 'Example <b>html</b>' (string) - A html string containing the rendered long_description_markdown
++ long_description_markdown: 'Example **markdown**' (string) - Markdown formatted string containing the project long description
++ open_posts_count (number) - Number of open posts created for the project
++ closed_posts_count (number) - Number of closed posts created for the project
```
joshsmith: Do you want example `number`s here at all?

blueprint/api.apib
```diff
@@ -842,10 +916,63 @@ When creating a user category, the authenticated user is assumed by the server s
...
 + id: `1` (string)
 + type: `projects` (string)
 
+## Project Attributes (object)
++ slug: 'example_slug' (string) - Case insensitive unique string identifier for the project
++ title: 'An example project' (string) - A title for the project
++ description: 'An example project description' (string) - A short plain text description for the project
++ icon_thumb_url: 'example.com/icon_thumb.png' (string) - A url pointing to the project thumbnail
++ icon_large_url: 'example.com/icon_large.png' (string) - A url pointing to a large icon for the project
++ long_description_body: 'Example <b>html</b>' (string) - A html string containing the rendered long_description_markdown
```
joshsmith, line 1736: "An html"

joshsmith:
Do you want to have `## Project` and `## Projects` sections like one of the other endpoints, or no?

joshsmith:
I think we should make it clear what is and is not valid in the `slug`.

joshsmith:
@begedin number of answers to your questions, some little notes here and some questions. If you make changes, please do commit without squashing.

PR #480, opened 2016-07-28
Add documentation for Organization Memberships API

### LINKED ISSUE #409: Document Organization Memberships API

begedin:
Closes #409

### DIFF
--- a/blueprint/api.apib
+++ b/blueprint/api.apib
@@ -148,6 +148,111 @@ Exchange credentials for `access_token`.
 
 # Group Organization Memberships
 
+## Organization Memberships [/organization_memberships]
+
+### Filter organization memberships by id [GET]
+
++ Parameters
+
+    + filter
+        + id: `1,2,...` (string, required) - Comma separated string of `ids` to filter by.
+
++ Headers
+
+    Authorization: Bearer {access_token}
+
++ Response 200 (application/vnd.api+json)
+
+    + Attributes (Organization Memberships Response)
+
+### Create an organization membership [POST]
+
++ Request
+
+    + Attributes (Organization Membership Create Request)
+
++ Headers
+
+    Authorization: Bearer {access_token}
+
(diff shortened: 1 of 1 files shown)


begedin:
@JoshSmith In several of my PRs, there's the debate of bundling up resources by "contains id, does not contain id". I think it may be worth doing, but the thing I'm unsure of, in this specific example, are cases like

``` Markdown
### List all organization memberships for an organization [GET /organizations/:id/memberships]
```

Where does that go? 

Do I keep it under 

```
## Organization Memberships [/organization_memberships]
```

Do I move it under the "Organizations" group?

I'll make it a point to update all my PRs, since I really did miss the grouping (mostly because the first endpoint being grouped are the categories and they seem to be completely wrong, using urls that do not exist, etc., but we do need to establish what to do with these, "special" endpoints, which do not follow the regular URL convention.

Also, what if a resource doesn't have the index/create actions and for example, only has those where an id needs to be specified? Do we still keep the grouping, or do we move it a level up?

blueprint/api.apib
```diff
@@ -148,6 +148,111 @@ Exchange credentials for `access_token`.
...
 # Group Organization Memberships
 
+## Organization Memberships [/organization_memberships]
+
+### Filter organization memberships by id [GET]
+
++ Parameters
+
+    + filter
+        + id: `1,2,...` (string, required) - Comma separated string of `ids` to filter by.
```
npendery, line 158: Shouldn't we use the `Id filter object` here?

blueprint/api.apib
```diff
@@ -148,6 +148,111 @@ Exchange credentials for `access_token`.
...
+
++ Headers
```
npendery, line 160: I might be using old formatting, but I think we want the header under a request as:

```
+ Request

  + Headers

       Accept: application/vnd.api+json

       Authorization: Bearer {access_token}
```

PR #495, opened 2016-08-06
Adding documentation for Post Like API

schippada:


schippada:
Please let me know if the response JSON is correct for the create request.

app/controllers/posts_controller.rb
```diff
@@ -24,6 +24,7 @@ class PostsController < ApplicationController
   def index
     authorize Post
     includes = [:user, :project]
+    id_params = params[:filter]
```
joshsmith: This line should be removed.

